PR #4, opened 2017-03-05
Feature/lw17 embed hslogger

volhovm:
Most important changes:
* `hslogger` was completely copied into log-warper, dependencies are removed.
* `Growl` and `Log4jXML` handlers were removed (are redundant).
* `System.Log` (that contained `Priority`) is removed, `System.Wlog.Severity#Severity` is used instead everywhere. 
* All messages are now `Text`. `LogRecord = (Severity, Text)` is now, it's located in `Severity` module as well. `Text.IO` output functions (e.g. `TIO.hPutStrLn`)is actively used instead of `IO` alternatives.
* `System.Log.Formatter` was merged into `System.Wlog.Formatter`. As a result, `setStdoutFormatter` and `setStderrFormatting` functions were removed from `System.Wlog.Formatter` and in places of usage you just do `setformatter stdoutFormatter ...` (the problem is cycle dependency of `Formatter` and `Handler`

What is to be done next (not in this issue though):
* Licensing issues and commentaries (make our own, add copyrights). I didn't follow all of them carefully.


log-warper.cabal
```diff
@@ -22,31 +22,27 @@ library
...
   other-modules:       System.Wlog.Color
-                       System.Wlog.Formatter
                        System.Wlog.MemoryQueue
 
   build-depends:       aeson                >= 0.11.2.1
                      , ansi-terminal        >= 0.6.2.3
                      , base                 >= 4.7 && < 5
                      , bytestring           >= 0.10.8.1
-                     , containers
+                     , containers           >= 0.5.7.0
```
chshersh, line 45: Maybe delete all version bounds? I put them in the beginning because I was stupid. Now I think that using `stack` it's more convenient to not use lower-bounds. And `stack` also can add correct bounds on uploading to `hackage`.

src/System/Wlog/Color.hs
```diff
@@ -11,29 +11,35 @@
...
     setColor color = setSGRCode [SetColor Foreground Vivid color]
     reset = setSGRCode [Reset]
 
--- | Colorizes text.
-colorizer :: Priority -> String -> String
-colorizer pr s = before ++ s ++ after
-  where
-    (before, after) = table pr
+-- | Colorizes string text.
+colorizer :: Severity -> String -> String
```
chshersh, line 38: For moment I thought that this coloring functionality should be nicely merged with `serokell-util/ANSI`: https://github.com/serokell/serokell-util/blob/af79ca0c42b8fd23c499096f91f649edfb0862f1/src/Serokell/Util/ANSI.hs#L16

But then I realized that we don't depend on  `serokell-util`. Maybe we should eventually?

src/System/Wlog/Handler/Simple.hs
```diff
@@ -0,0 +1,100 @@
+{-# LANGUAGE CPP               #-}
+{-# LANGUAGE NoImplicitPrelude #-}
```
chshersh, line 2: You decided it's not worth it to use `Universum` in every module?)

src/System/Wlog/Color.hs
volhovm: I still don't think that importing `serokell-util` is good, because it (once again):
1. Has a lot of dependencies
2. Lot of them are not in any LTS
3. It's not in LTS itself.

I find copiyng small fragments of libraries a good decision in some cases, this one counts.

log-warper.cabal
volhovm: I don't know anything about how `stack` does that. Still think it's better to have stack as secondary build tool because anyway we define a lot of building settings in `.cabal` file. You can try to reassure me (in slack) if you find my pov wrong.

src/System/Wlog/Handler/Simple.hs
volhovm: We depend on it already, so anyway :)

PR #5, opened 2017-03-15
[LW-19] Add features to decrease boilerplate

Martoon-00:


log-warper.cabal
```diff
@@ -1,5 +1,5 @@
 name:                log-warper
-version:             1.0.1
+version:             1.0.4
```
gromakovsky: Why does 1.0.4 follow 1.0.1? 🤔

gromakovsky: Hm, ok, I see that haddock version is currently 1.0.3, but it's strange.

Martoon-00: I forgot to pull master with +0.0.2 to ver. before creating this branch (￣ェ￣;)

PR #9, opened 2017-04-05
Feature/csl837 fake logging revised

volhovm:
* Added more decent handler tags `Filelike` or `Other`, see `Handler.hs`
* Moved time rounding into handler (`LoggerConfig`), it's not a global variable anymore. Also changed `FromJSON` instance a little bit -- now instead of `file` or `files` you can specify `handlers`. 
* Moved `Roller` into `Handler.Roller`. 
* Added `readBack` into `LogHandler` class and implemented it: for `GenericHandler` messages are kept using `MemoryQueue`, for `RollerHandler` back-reading is implementing by taking `MVar Handle`, reading from it and re-opening.


src/System/Wlog/Handler.hs
```diff
@@ -31,7 +31,10 @@ import           System.Wlog.Severity  (LogRecord, Severity)
 import           Universum
 
 -- | Tag identifying handlers.
-data LogHandlerTag = HandlerFilelike FilePath | HandlerOther String
+data LogHandlerTag
```
chshersh: Could you clarify more in documentation what is the difference between `HandlerFileLike` and `HandlerOther`?

chshersh (approved):

PR #11, opened 2017-04-22
[CSL-610,979] Add 'HasLoggerName Identity' instance

rockbmb:
@ChShersh can you merge this ASAP? I know I should've just asked to push directly to `master` but this was faster since I made this branch a few days ago.

### DIFF
--- a/src/System/Wlog/LoggerNameBox.hs
+++ b/src/System/Wlog/LoggerNameBox.hs
@@ -78,6 +78,11 @@ instance (Monad m, HasLoggerName m, Monoid w) => HasLoggerName (RWSStrict.RWST r
 
     modifyLoggerName = RWSStrict.mapRWST . modifyLoggerName
 
+instance HasLoggerName Identity where
+    getLoggerName = Identity mempty
+
+    modifyLoggerName = flip const
+
 -- | Set logger name in context.
 setLoggerName :: HasLoggerName m => LoggerName -> m a -> m a
 setLoggerName = modifyLoggerName . const


src/System/Wlog/LoggerNameBox.hs
```diff
@@ -78,6 +78,11 @@ instance (Monad m, HasLoggerName m, Monoid w) => HasLoggerName (RWSStrict.RWST r
 
     modifyLoggerName = RWSStrict.mapRWST . modifyLoggerName
 
+instance HasLoggerName Identity where
+    getLoggerName = mempty
```
gromakovsky: I would use some meaningful name here, at least `Identity`. Otherwise someone can accidentally use this instance and not understand what's going on.

rockbmb:
@ChShersh I don't understand this part:
> modifyLoggerName could be `<>`

`modifyLoggerName` has type `(LoggerName -> LoggerName) -> m a -> m a`, can you explain how it can be `<>`?

chshersh:
@rockbmb Oh, sorry, I forgot type of `modifyLoggerName`. You're right, it cannot be `<>`.

rockbmb:
@ChShersh in that case, the only change you want me to make is `Identity "identity"`. I'll do this right now.

gromakovsky:
Well, if you think that it should be `mempty` instead of `"Identity"`, you can try to convince me. I wrote one argument why I think it should be `"Identity"`.

rockbmb:
@ChShersh can you do that for me? I just tried to push to `master` and remembered I don't have permissions, and a PR for a single line is time wasterful.

rockbmb:
@gromakovsky I should have said something indicating my agreement, but I agree with you. `mempty` is not helpful, and although we might not need this instance, it might be helpful in the future as it has a little bit more information that just `""`.

PR #16, opened 2017-09-15
Support Unix paths in log configs even on Windows

neongreen:
This is needed for https://issues.serokell.io/issue/CSL-1591

### DIFF
--- a/src/System/Wlog/LoggerConfig.hs
+++ b/src/System/Wlog/LoggerConfig.hs
@@ -68,6 +68,7 @@ import           Data.Yaml              (FromJSON (..), ToJSON (..), Value (Obje
                                          object, (.!=), (.:), (.:?), (.=))
 import           Formatting             (bprint, shown)
 import           GHC.Generics           (Generic)
+import           System.FilePath        (normalise)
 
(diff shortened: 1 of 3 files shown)


src/System/Wlog/LoggerConfig.hs
```diff
@@ -122,8 +123,9 @@ instance Monoid LoggerTree where
 
 instance ToJSON HandlerWrap
 instance FromJSON HandlerWrap where
+    -- we use 'normalise' so that Unix paths would work on Windows
     parseJSON = withObject "handler wrap" $ \o -> do
-        (_hwFilePath :: FilePath) <- o .: "file"
+        (_hwFilePath :: FilePath) <- normalise <$> o .: "file"
```
neongreen, line 128: http://hackage.haskell.org/package/filepath-1.4.1.2/docs/System-FilePath-Windows.html#v:normalise says that it converts `/` to `pathSeparator`, so this should correctly convert relative Unix paths (such as `pub/node.pub`) to Windows paths on Windows

neongreen:
Misha okayed it sort of

PR #17, opened 2017-09-25
[CSL-1703] Fix space leak in readback buffer

avieth:
```
{-# LANGUAGE OverloadedStrings #-}

import System.Wlog

main = do
  parseLoggerConfig "log-config.yaml" >>= setupLogging
  usingLoggerName "node" logForever

logForever :: LoggerNameBox IO ()
logForever = do
    logDebug "debug"
    logForever
```

```
# log-config.yaml
severity: Info
node:
    severity: Debug
```

This program would leak memory unless the read back buffer memory queue is regularly forced by some other means. With this patch it won't leak past the bounds given (`2 * 1024 * 1024` currently). Should mention as well that the limit is actually on the number of characters in the memory queue, rather than the number of bytes, so docs should be updated.

### DIFF
--- a/src/System/Wlog/Handler/Simple.hs
+++ b/src/System/Wlog/Handler/Simple.hs
@@ -69,7 +69,9 @@ streamHandler h sev = do
     mq <- newMVar $ MQ.newMemoryQueue $ 2 * 1024 * 1024 -- 2 MB
     let mywritefunc hdl msg = withMVar lock $ const $ do
             writeToHandle hdl msg
-            modifyMVar_ mq $ pure . pushFront msg
+            -- Important to force the queue here, else a massive closure will
+            -- be retained until the queue is actually used.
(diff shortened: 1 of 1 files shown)


chshersh (approved):

chshersh:
@volhovM Could you please review and approve as well?

adinapoli-iohk:
@avieth I haven't profiled anything in that area, but wouldn't [this line](https://github.com/serokell/log-warper/pull/17/files#diff-7b86c07e3eb88ac12fab0a935cd17e4bR58) suffer a similar malady?

volhovm (approved):

avieth:
> @avieth I haven't profiled anything in that area, but wouldn't this line suffer a similar malady?

As we agreed, it's not so serious: the user of `readBack` will probably force the list right away, and it's not putting un-forced data into a mutable cell. Still, I'll be defensive and make it strict.

dcoutts (approved):
LGTM.

PR #19, opened 2017-09-27
Improve memory efficiency

adinapoli-iohk:
This beefy PR improves the memory allocation of the library, turning it from exponential/linear to constant in the majority of cases (_caveat emptor_, which are explained below). Below we move to the explanation, let me give you the 10.000ft overview of what this patch provides. This is explained in detail below:

* The `MemoryQueue` has been partially reworked to get rid of the "inline" State manipulation;
* Strings have been dropped almost everywhere in favour of `Text`;
* A `LogFormatter` has been reworked to yield a `IO Builder`;
* `replaceVarM` has been reworked to be pure _and_ to work with builders rather than plain Text/Strings;
* The pure logger has been reworked to use strict's `StateT` instead of WriterT;
* The pure logger have been polished to drop instances which required the `UndecidableInstances` pragma;
* The `Sized` instance for `Text` has been reworked and multiplied by a constant factor of 16 (see below).

## The thought process

To explain why I did what I did, let's start with a simple program I have whipped up, basically cloning `examples/Playground.hs`:

```haskell
{-# LANGUAGE NoImplicitPrelude   #-}
{-# LANGUAGE OverloadedStrings   #-}
{-# LANGUAGE ScopedTypeVariables #-}
-- | Tests for the Formatter module.

module Main where

import           Universum

import           Data.Monoid      ((<>))
import           Data.Yaml.Pretty (defConfig, encodePretty)

import           System.Wlog      (CanLog, buildAndSetupYamlLogging,
                                   dispatchEvents, logDebug, logError, logInfo,
                                   logNotice, logWarning, modifyLoggerName,
                                   parseLoggerConfig, prefixB, productionB,
                                   releaseAllHandlers, runPureLog,
                                   usingLoggerName)

testLoggerConfigPath :: FilePath
testLoggerConfigPath = "logger-config-example.yaml"

testToJsonConfigOutput :: MonadIO m => m ()
testToJsonConfigOutput = do
    cfg             <- parseLoggerConfig testLoggerConfigPath
    let builtConfig  = cfg <> productionB <> prefixB "logs"
    Universum.putStrLn $ encodePretty defConfig builtConfig

testLogging :: (CanLog m) => Int -> m ()
testLogging runNum = usingLoggerName "node" $ do
    logDebug   $ "skovoroda" <> Universum.show runNum
    logInfo    "patak"
    logNotice  "boroda"
    logWarning "haha"

    modifyLoggerName (<> "server") $ do
        logDebug  "provoda"
        logNotice "Ggurda"

    logError   "BARDAQ"

showPureLog :: Int -> IO ()
showPureLog idx = do
    (res, pureLog) <- runPureLog (testLogging idx)
    putText $ "Pure log: " <> fromString (Universum.show idx)
    usingLoggerName "naked" $ do
        logWarning $ "Pure log for result = " <> Universum.show res <> ":"
        dispatchEvents pureLog

main :: IO ()
main = do
    testToJsonConfigOutput
    bracket_
        (buildAndSetupYamlLogging (productionB <> prefixB "logs") testLoggerConfigPath)
        releaseAllHandlers
        doWork
  where
    doWork = forM_ [0.. 10000] $ \idx -> do
      testLogging idx
      showPureLog idx
```

In a nutshell, we log stuff 10k times and we exit. When I started yesterday, the memory allocation was pretty much like this (**Note: some numbers might feel inconsistent as I have varied in some cases the number of iterations or the size of the memory queue, but the allocation patter is what matters.**):

screen shot 2017-09-27 at 10 44 50

As you can see the memory was growing steadily but slowly. In particular, this chart reveals we were allocating the most in:

* `replaceVarM`
* `modify/pushFront`

I decided to tackle the latter first. The code GHC is referring to is the following:

```haskell
pushFront :: (Sized a) => a -> MemoryQueue a -> MemoryQueue a
pushFront msg mq = executingState mq $ do
    mqMemSize += getSize msg
    mqQueue   %= (msg <|)
    whileM_ isLimitExceeded $
        modify (snd . popLast)
  where
    isLimitExceeded = liftA2 (<) (use mqLimit) (use mqMemSize)
```

The problem with this code are the numerous call to `modify`, both in the form of the function and also using `(%=)`, which alas is using the lazy version of `modify` under the hood. Let's also remember that using strict `State` is something not _enough_, because that ensures only that sequencing of actions are strict, _not_ that the overall state `s` is, so we can still leak. I ended up rewriting the function to basically "unroll" the State. After I did that, `modify` disappeared from the chart (alas I don't have an intermediate chart to show).

In order to tackle `replaceVarM` I acted on two fronts (as I was coming from numerous failed attempts): I did make the function pure, because the only reason why we were doing `IO` there seems to have been the fact we wanted to grab the current time very precisely, but in reality no harm is done if we do this inside a `LogFormatter` as the delay won't be even noticeable.

After doing that I realised that the majority of the allocation was dominated by `append`, which is coming from `Data.Text.Array` and is very likely how `mappend` is implemented for two `Text` fragments. In other words, we were doing a lot of mappends inside `replaceVarM`, so my solution was to use a `Builder` instead so that those became O(1).

After I did that, the chart started to look better but not super:

screen shot 2017-09-27 at 14 11 12

At this point, I think I was forcing the Builder into a `Text` before we were passing it to the `emit` function, so the next step was to delay as much as possible the evaluation by pushing that at the very bottom of the evaluation chain.

Then I moved to the pure logger, for which I replaced WriterT in favour of a strict StateT monad, for the classic reasons which can be found online all over the internet. I have also removed a bunch of derived instances which were missing the type variable and thus were requiring the `UndecidableInstances` pragma to be added, which I guess it makes no harm to remove (unless we are heavily relying on those instances, in which case I can add those back).

Last but not least, I have played with the `Sized` instance for `Text` and this is the most mysterious part, if you want. This PR keeps the memoryQueue maximum size untouched (2MB) but it tweak how much a text fragment occupies into memory. What I have realised is that probably the previous instance, being defined on the _length_ of the `Text` was allowing 2 millions chars to be stored into the Seq, which can be problematic if the queue is never demanded. This is the part of the library I'm still trying to grap completely, but I have noticed that changing the max size of the queue _and/or_ the `Sized` instance for Text the memory pattern changes. There are some example runs:

screen shot 2017-09-27 at 17 23 50
screen shot 2017-09-27 at 17 28 25
screen shot 2017-09-27 at 17 38 08
screen shot 2017-09-27 at 17 55 55
screen shot 2017-09-27 at 17 59 09

**This PR, with these settings, delivers the last 2 memory patterns**.

If you take a look at the retainers, it's almost always the `pushFront` and `popBack`, so my hunch is that we should make some params externally-configurable so that we can tweak the memory allocation in application code. It seems to me that there is quite a bit of contention in the memoryqueue as the limit grows and the `Sized` value reduces. Truth to be told, the current `Sized Text` instance is totally arbitrary, or originally I did multiple by 16 by getting the math wrong (I mentally mapped that figure to _bits_ rather than _bytes), but I soon realised that this affects quite a bit the allocation. I wouldn't be surprised if we need a better data structure or more strictness when needed.

Phew, that was a long-winded explanation, sorry about that. But hopefully by now it should be clearer what I did.

Let me know what you think!


rockbmb (approved):
Nothing out of the ordinary, LGTM. However, it'd be good if @ChShersh took a look before I merge.

adinapoli-iohk:
@rockbmb Something we should check (I was going to do this today) is to see if we are still logging on disk as we are expecting. I did run all my experiments and checked we do log on stdout, but it would be very lame to discover the library is now not logging as it should on disk.

Let me quickly confirm this as well. (ps. if we don't have such kind of tests, we should probably create a bunch of [golden tests](https://kseo.github.io/posts/2016-12-15-golden-tests-are-tasty.html) to catch these regressions).

adinapoli-iohk:
Ok, that's all good. I did get suspicious because I was using the example yaml file which was logging only 256 bytes worth of logs before rotating them, and I feared it was a regression.

chshersh:
@adinapoli-iohk Thanks for such great work! I finally reviewed your changes. This looks really amazing. Didn't notice any problems in code. But idea with having golden-tests for file writing (and having more tests at all) is really a nice idea :)

adinapoli-iohk:
Thanks @ChShersh ! That's really appreciated 😉 

> But idea with having golden-tests for file writing (and having more tests at all) is really a nice idea :)

I have stoled this great idea from the [katip](https://github.com/Soostone/katip) project (of which I am a great fan), and from all the awesome work @MichaelXavier and the Soostone guys did on it 😀

adinapoli-iohk:
@ChShersh @neongreen @gromakovsky Follow-up question for you: Would it be possible to release a version of `log-warper` which ships only the memory improvements but _not_ the API breaking-changes introduced by 1.3.0, to ease the adoption into upstream packages? Thank you!

chshersh:
@adinapoli-iohk I think that's possible. I can try to do that.

chshersh:
@adinapoli-iohk Here is it is: http://hackage.haskell.org/package/log-warper-1.2.3.1
I've checked out to this commit https://github.com/serokell/log-warper/commit/24f92d0f05c2bfe812857db71f0210f2f65f0e3d and released it under version `1.2.3.1`

adinapoli-iohk:
Awesome @ChShersh , thanks a ton! 🎉

PR #27, opened 2017-10-14
[CSL-1769] Remove handler from root logger

volhovm:


chshersh:
@volhovM Could you please explain motivation behind this change?

volhovm:
Without it, even if `initTerminalLogging` is not called, terminal logging is enabled (but is terribly ugly). Which doesn't make sense.

chshersh (approved):

chshersh:
Bump up version and upload to Hackage :)

chshersh:
You can even write to CHANGELOG.md

int-index (approved):
I tested this patch and it fixes the issue I had in CSL.

PR #28, opened 2017-10-16
[CSL-1778] correct logger config parsing

avieth:
Terminal logging severity couldn't be set due to a missing top level key.


chshersh:
@avieth Thanks a lot for that fix! I wish something more clever can be used to prevent such bugs.

chshersh (approved):

avieth:
> I wish something more clever can be used to prevent such bugs.

We could change the format of the logger config yaml so that the the top-level configuration keys don't get mixed up with the named logger trees: put all of the logger tree keys in an object under the key `tree`, for instance.

chshersh:
Agree. That would be a great change.

PR #37, opened 2017-11-05
vrom911/LW-27 fix TODO

vrom911:


src/System/Wlog/Logger.hs
```diff
@@ -81,15 +80,15 @@ data HandlerT = forall a. LogHandler a => HandlerT a
 data Logger = Logger
     { _lLevel    :: Maybe Severity
     , _lHandlers :: [HandlerT]
-    , _lName     :: String
+    , _lName     :: Text
```
chshersh: Can we also use `LoggerName` newtype here?

src/System/Wlog/Logger.hs
```diff
@@ -81,15 +80,15 @@ data HandlerT = forall a. LogHandler a => HandlerT a
...
     } deriving (Generic)
 
 makeLenses ''Logger
 
-type LogTree = Map String Logger
+type LogTree = Map Text Logger
```
chshersh: And here.

src/System/Wlog/Logger.hs
```diff
@@ -81,15 +80,15 @@ data HandlerT = forall a. LogHandler a => HandlerT a
...
 
 data LogInternalState = LogInternalState
-    { liTree   :: Map String Logger
+    { liTree   :: Map Text Logger
```
chshersh: And replace `Map Text Logger` with `LogTree`.

src/System/Wlog/Logger.hs
```diff
@@ -124,30 +123,30 @@ Example return value:
...
+        let newlevel = accum <> "." <> x
         in newlevel : joinComp xs newlevel
 
 ---------------------------------------------------------------------------
 -- Logging With Location
 ---------------------------------------------------------------------------
 
 -- | Log a message using the given logger at a given priority.
-logM :: String     -- ^ Name of the logger to use
+logM :: Text       -- ^ Name of the logger to use
```
chshersh: And here. Basically, it's really good idea to use `LoggerName` newtype everywhere you intend to use `LoggerName`. So it won't be so easy to mess up with function arguments.

src/System/Wlog/LoggerNameBox.hs
```diff
@@ -39,52 +39,52 @@ import           System.Wlog.LoggerName      (LoggerName)
 -- TODO: replace current methods with Lens?
 class HasLoggerName m where
     -- | Extract logger name from context
-    getLoggerName :: m LoggerName
+    askLoggerName :: m LoggerName
```
chshersh: Could you please add `default` implementations of `askLoggerName` and `modifyLoggerName` (using `hoist` from `MFunctor`? This will help to remove boilerplate.

Just see how it's done here for `CanLog` type class:

https://github.com/serokell/log-warper/blob/ec162c6557ec25939c7fa37bfaaa3c4a0682ba5a/src/System/Wlog/CanLog.hs#L70

CHANGES.md
```diff
@@ -1,3 +1,9 @@
+1.5.0
+=====
+
+* Replace `String` to `Text` in `LoggerName`
+* Rename `LoggerName` field name to `getLoggerName`
```
chshersh: Also you forgot to mention that you renamed `getLoggerName` inside `HasLoggerName` to `askLoggerName`.

chshersh (changes requested):

src/System/Wlog/LoggerNameBox.hs
```diff
@@ -44,6 +45,15 @@ class HasLoggerName m where
...
+    askLoggerName = lift askLoggerName
+
+    default modifyLoggerName :: (MFunctor t, t n ~ m, Monad n, HasLoggerName n)
+                             => (LoggerName -> LoggerName)
+                             -> m a
+                             -> m a
+    modifyLoggerName f = hoist (modifyLoggerName f)
+
 instance (Monad m, HasLoggerName m) => HasLoggerName (ReaderT a m) where
     askLoggerName = lift askLoggerName
```
chshersh: You should also remove all bodies of all boilerplate instances for `HasLoggerName` type class. The point of having default signatures is too make all instances shorter as possible and remove all boilerplate. In that way you can also verify that your default implementations are indeed correct. Maybe only for `ContT` monad your default signature won't work... But that's okay, usually nothing works for `ContT` and this monad should be avoided.

src/System/Wlog/Logger.hs
```diff
@@ -98,8 +100,8 @@ data LogInternalState = LogInternalState
 
 -- | The name of the root logger, which is always defined and present
 -- on the system.
-rootLoggerName :: Text
-rootLoggerName = ""
+rootLoggerName :: LoggerName
+rootLoggerName = LoggerName ""
```
chshersh: I would like to define it like this: `rootLoggerName = mempty`.

chshersh (changes requested):

src/System/Wlog/LoggerNameBox.hs
vrom911: Sorry, forgot about that.

vrom911: `ContT` monad is not working with default implementations, as you expected, so I left it as it is.

chshersh:
@vrom911 Okay, that's really good! One last shy request: could you please move `HasLoggerName` type class with all boilerplate instances into separate module `System.Wlog.HashLoggerName` and move all pure logging into separate module `System.Wlog.PureLogging`? You already performed big refactoring so let's do this as well! The idea is to have each entity in its own module: type class in separate module, `LoggerNameBox` in separate module, `CanLog` type class in separate module and whole pure logging system inside separate module.

src/System/Wlog/#LoggerName.hs#
```diff
@@ -0,0 +1,19 @@
+{-# LANGUAGE TemplateHaskell #-}
```
chshersh: This file is redundant here. You probably need to add `#*` to `.gitignore`

chshersh (changes requested):

chshersh (approved):

Issue #38, opened 2017-11-07
Add `usingNamedPureLogger` to pure logging


vrom911:
See https://github.com/serokell/ale-core/pull/36#discussion_r149497536

vrom911:
PR #39

PR #39, opened 2017-11-07
[#38] Add usingNamedPureLogger function

vrom911:


src/System/Wlog/PureLogging.hs
```diff
@@ -123,3 +124,12 @@ launchNamedPureLog hoist' (NamedPureLogger action) = do
     name <- askLoggerName
     (logs, res) <- hoist' $ swap <$> usingLoggerName name (runPureLog action)
     res <$ dispatchEvents logs
+
+-- | Similar to 'runNamedPureLog', but using provided logger name.
+usingPureLoggerName
+    :: Monad m
```
chshersh: I think you can remove `Monad` constraint here and from `runPureLog` function as well. Or at least you can relax it to `Functor`...

chshersh (changes requested):

chshersh (approved):

PR #41, opened 2017-11-08
[#40] Add 'logPureAction' function

### LINKED ISSUE #40: Add `logPureAction` to pure logging

vrom911:
See https://github.com/serokell/ale-core/pull/36#issuecomment-342656946
Resolves #40

### DIFF
--- a/src/System/Wlog/PureLogging.hs
+++ b/src/System/Wlog/PureLogging.hs
@@ -20,7 +20,8 @@ module System.Wlog.PureLogging
        , NamedPureLogger (..)
        , runNamedPureLog
        , launchNamedPureLog
-       , usingPureLoggerName
+       , usingNamedPureLogger
+       , logPureAction
        ) where
 
 import           Universum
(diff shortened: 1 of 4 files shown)


src/System/Wlog/PureLogging.hs
```diff
@@ -126,9 +127,18 @@ launchNamedPureLog hoist' (NamedPureLogger action) = do
...
     usingLoggerName name $ runPureLog action
+
+-- | Perform pure-logging computation, log its events
+-- and return the result of the computation
+logPureAction :: WithLogger m => NamedPureLogger m a -> m a
+logPureAction namedPureLogger = do
+    loggerName    <- askLoggerName
+    (a, logEvnts) <- usingNamedPureLogger loggerName namedPureLogger
+    logEvents logEvnts
+    pure a
```
chshersh: This can be written in shorter form `a <$ dispatchEvents eventsBacklog`.

I also think that it's better to use `dispatchEvents` here instead of `logEvents` because you already have `LogEvent`s with proper `LoggerName`. You don't need to use `askLoggerName` (inside `logEvents`) second time.

src/System/Wlog/PureLogging.hs
```diff
@@ -126,9 +127,18 @@ launchNamedPureLog hoist' (NamedPureLogger action) = do
...
-usingPureLoggerName :: Functor m
-                    => LoggerName
-                    -> NamedPureLogger m a
-                    -> m (a, [LogEvent])
-usingPureLoggerName name (NamedPureLogger action) =
+usingNamedPureLogger :: Functor m
+                     => LoggerName
+                     -> NamedPureLogger m a
+                     -> m (a, [LogEvent])
+usingNamedPureLogger name (NamedPureLogger action) =
```
chshersh, line 134: Thanks for renaming this function! This is much better for naming convention :)

chshersh (changes requested):
@vrom911 Could you also please add function `withSublogge name = modifyLoggerName (<> name)`? This function is really helpful and convenient. So let's just add it to this PR, I don't want to create separate PR for something simple...

vrom911:
@ChShersh I've added `withSublogger` to the `HasLoggerName` module. Tell me if you disagree with the placement.

chshersh (approved):

PR #43, opened 2017-11-08
[#42] Implement launchNamedPureLogWith

### LINKED ISSUE #42: Implement `launchPureLogWithPure` function
See https://github.com/serokell/ale-core/pull/36#discussion_r149788604

vrom911:
Resolves #42

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+1.5.3
+=====
+
+* Add `launchNamedPureLogWith` to `PureLogging`
+* Improve documentation for `launchNamedPureLog`
+
 1.5.2
 =====
 
--- a/src/System/Wlog/PureLogging.hs
+++ b/src/System/Wlog/PureLogging.hs
@@ -20,6 +20,7 @@ module System.Wlog.PureLogging
        , NamedPureLogger (..)
        , runNamedPureLog
        , launchNamedPureLog
+       , launchNamedPureLogWith
        , usingNamedPureLogger
        , logPureAction
        ) where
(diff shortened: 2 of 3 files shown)


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+1.5.3
+=====
+
+* Add `launchNamedPureLogWithPure` to `PureLogging`
+* Improve documentation for `launchPureLog`
```
chshersh: I think you want to say that you improved documentation for `launchNamedPureLog` function :)

src/System/Wlog/PureLogging.hs
```diff
@@ -115,17 +116,50 @@ runNamedPureLog
...
+{- | Similar to 'launchPureLog', but provides logger name from current context.
+
+Running the 'NamedPureLogger' gives us the pair  of target and the list of 'LogEvent's,
+wrapped in 'Monad' from where using the fact that @(,)@ is 'Functor' logging can be triggered.
+
+==== __Example__
+
+@
+  newtype PureSmth a = ...
+      deriving (MonadSmth, ...)
```
chshersh, line 128: I think example also should contain string `Instance MonadSmth m => MonadSmt (NamedLoggerName m)` or something like this. I guess without such instance it won't work...

src/System/Wlog/PureLogging.hs
```diff
@@ -115,17 +116,50 @@ runNamedPureLog
...
+launchNamedPureLog hoist' namedPureLogger = do
     name <- askLoggerName
-    (logs, res) <- hoist' $ swap <$> usingLoggerName name (runPureLog action)
+    (logs, res) <- hoist' $ swap <$> usingNamedPureLogger name namedPureLogger
     res <$ dispatchEvents logs
 
+-- | Similar to 'launchNamedPureLog', but lift the value of function argument.
+launchNamedPureLogWithPure
+    :: (WithLogger n, Monad m)
+    => (forall f. Functor f => m (f a) -> (f b))
```
chshersh: `()` around `f b` are redundant.

src/System/Wlog/PureLogging.hs
```diff
@@ -115,17 +116,50 @@ runNamedPureLog
...
+  evalPureSmth :: PureSmth a -> a
+
+  makeField    :: MonadSmth m => Data -> m Field
+
+  run :: (MonadIO m, WithLogger m) => m ()
+  run = do
+      data  <- getData
+      -- field :: Field
+      field <- launchNamedPureLog (pure . evalPureSmth)
+                                  (makeField data)
```
chshersh: I think this function call can be written in one line. It's not very long.

src/System/Wlog/PureLogging.hs
```diff
@@ -115,17 +116,50 @@ runNamedPureLog
...
     -> NamedPureLogger m a
     -> n b
-launchNamedPureLog hoist' (NamedPureLogger action) = do
+launchNamedPureLog hoist' namedPureLogger = do
     name <- askLoggerName
-    (logs, res) <- hoist' $ swap <$> usingLoggerName name (runPureLog action)
+    (logs, res) <- hoist' $ swap <$> usingNamedPureLogger name namedPureLogger
     res <$ dispatchEvents logs
 
+-- | Similar to 'launchNamedPureLog', but lift the value of function argument.
```
chshersh: I think it's better to say things explicitly, like `but call 'pure' on passed function result. Also it's good idea to to add difference with the `launchNamedPureLog` like tell that this should be called like 

```haskell
field <- launchNamedPureLogWithPure evalPureSmth $ makeField data
```

src/System/Wlog/PureLogging.hs
```diff
@@ -115,17 +116,50 @@ runNamedPureLog
...
+launchNamedPureLogWithPure
```
chshersh: Yeah, this name is not really good... :)
But I can't think of really better name. Maybe `Pure` part is redundant and we can call it just `launchNamedPureLogWith`.

chshersh (changes requested):
This PR is really good! But some cosmetic changes are required.

chshersh (approved):

PR #44, opened 2017-11-11
[LW-28] Remove duplicate printing of ERROR

### LINKED ISSUE #45: Get rid of duplicate error log printing

vrom911:
Resolves #45

### DIFF
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+
 -- | Testing module to play with logging.
 
 module Main where
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -1,19 +1,16 @@
-{-# LANGUAGE NoImplicitPrelude #-}
-
 -- | Testing module to play with logging.
 
 module Main where
 
-import           Universum
+import Universum
 
-import           Data.Monoid      ((<>))
-import           Data.Yaml.Pretty (defConfig, encodePretty)
+import Data.Monoid ((<>))
+import Data.Yaml.Pretty (defConfig, encodePretty)
 
-import           System.Wlog      (CanLog, Severity (Debug), buildAndSetupYamlLogging,
-                                   dispatchEvents, logDebug, logError, logInfo, logNotice,
-                                   logWarning, modifyLoggerName, parseLoggerConfig,
-                                   prefixB, productionB, releaseAllHandlers, runPureLog,
-                                   termSeverityB, usingLoggerName)
+import System.Wlog (CanLog, Severity (Debug), buildAndSetupYamlLogging, dispatchEvents, logDebug,
+                    logError, logInfo, logNotice, logWarning, modifyLoggerName, parseLoggerConfig,
+                    prefixB, productionB, removeAllHandlers, runPureLog, termSeverityB,
+                    usingLoggerName)
 
 testLoggerConfigPath :: FilePath
 testLoggerConfigPath = "logger-config-example.yaml"
@@ -51,5 +48,5 @@ main = do
     let config = (productionB <> prefixB "logs" <> termSeverityB Debug)
     bracket_
         (buildAndSetupYamlLogging config testLoggerConfigPath)
-        releaseAllHandlers
+        removeAllHandlers
         (testLogging >> showPureLog)
--- a/src/System/Wlog/CanLog.hs
+++ b/src/System/Wlog/CanLog.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE DefaultSignatures     #-}
 {-# LANGUAGE FlexibleInstances     #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE NoImplicitPrelude     #-}
 {-# LANGUAGE OverloadedLists       #-}
 {-# LANGUAGE Rank2Types            #-}
 {-# LANGUAGE TypeFamilies          #-}
@@ -22,19 +21,20 @@ module System.Wlog.CanLog
        , logMessage
        ) where
 
-import           Universum
+import Universum
 
-import           Control.Monad.Except      (ExceptT)
-import qualified Control.Monad.RWS         as RWSLazy
-import qualified Control.Monad.RWS.Strict  as RWSStrict
-import qualified Control.Monad.State.Lazy  as StateLazy
-import           Control.Monad.Trans       (MonadTrans (lift))
+import Control.Monad.Except (ExceptT)
+import Control.Monad.Trans (MonadTrans (lift))
 
-import           System.Wlog.HasLoggerName (HasLoggerName (..))
-import           System.Wlog.IOLogger      (logM)
-import           System.Wlog.LoggerName    (LoggerName (..))
-import           System.Wlog.LoggerNameBox (LoggerNameBox (..))
-import           System.Wlog.Severity      (Severity (..))
+import System.Wlog.HasLoggerName (HasLoggerName (..))
+import System.Wlog.IOLogger (logM)
+import System.Wlog.LoggerName (LoggerName (..))
+import System.Wlog.LoggerNameBox (LoggerNameBox (..))
+import System.Wlog.Severity (Severity (..))
+
+import qualified Control.Monad.RWS as RWSLazy
+import qualified Control.Monad.RWS.Strict as RWSStrict
+import qualified Control.Monad.State.Lazy as StateLazy
 
 
 -- | Type alias for constraints 'CanLog' and 'HasLoggerName'.
@@ -56,7 +56,7 @@ class Monad m => CanLog m where
     dispatchMessage name sev t = lift $ dispatchMessage name sev t
 
 instance CanLog IO where
-    dispatchMessage name prior msg = logM name prior msg
+    dispatchMessage = logM
 
 instance CanLog m => CanLog (LoggerNameBox m)
 instance CanLog m => CanLog (ReaderT r m)
--- a/src/System/Wlog/CanLog.hs
+++ b/src/System/Wlog/CanLog.hs
@@ -31,7 +31,7 @@ import qualified Control.Monad.State.Lazy  as StateLazy
 import           Control.Monad.Trans       (MonadTrans (lift))
 
 import           System.Wlog.HasLoggerName (HasLoggerName (..))
-import           System.Wlog.Logger        (logM)
+import           System.Wlog.IOLogger      (logM)
 import           System.Wlog.LoggerName    (LoggerName (..))
 import           System.Wlog.LoggerNameBox (LoggerNameBox (..))
 import           System.Wlog.Severity      (Severity (..))
--- a/src/System/Wlog/Handler.hs
+++ b/src/System/Wlog/Handler.hs
@@ -32,7 +32,7 @@ import           Data.Text.Lazy.Builder as B
 
 import           System.Wlog.Formatter  (LogFormatter, nullFormatter)
 import           System.Wlog.LoggerName (LoggerName (..))
-import           System.Wlog.Severity   (LogRecord (..), Severity)
+import           System.Wlog.Severity   (LogRecord (..), Severity (Error))
 
 -- | Tag identifying handlers.
 data LogHandlerTag
@@ -54,6 +54,9 @@ class LogHandler a where
     -- | Gets the current level.
     getLevel :: a -> Severity
 
+    -- | Needed for printing 'Error' only to @stderr@
+    shouldPrintError :: a -> Bool
+
     -- | Set a log formatter to customize the log format for this Handler
     setFormatter :: a -> LogFormatter a -> a
     getFormatter :: a -> LogFormatter a
@@ -63,10 +66,11 @@ class LogHandler a where
     -- given by the most recent call to 'setLevel'.
     handle :: a -> LogRecord -> LoggerName -> IO ()
     handle h lr@(LR pri _) logname =
-        when (pri >= (getLevel h)) $ do
-            let lName = getLoggerName logname
-            formattedMsg <- (getFormatter h) h lr lName
-            emit h formattedMsg lName
+        when (pri >= (getLevel h)) $
+            when (not $ pri == Error && not (shouldPrintError h)) $ do
+                let lName = getLoggerName logname
+                formattedMsg <- (getFormatter h) h lr lName
+                emit h formattedMsg lName
 
     -- | Forces an event to be logged regardless of
     -- the configured level.
--- a/src/System/Wlog/IOLogger.hs
+++ b/src/System/Wlog/IOLogger.hs
@@ -43,7 +43,7 @@ module System.Wlog.IOLogger
        , addHandler, removeHandler, setHandlers
        , getLevel, setLevel, clearLevel
          -- ** Severity settings
-       , setSeverity, setSeverityMaybe
+       , setSeverities, setSeveritiesMaybe
          -- ** Saving Your Changes
        , saveGlobalLogger
        , updateGlobalLogger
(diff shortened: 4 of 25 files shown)


src/System/Wlog/Wrapper.hs
```diff
@@ -40,10 +40,11 @@ import           System.Wlog.Severity       (Severity (..))
 streamHandlerWithLock :: (Handle -> Text -> IO ())
                       -> MVar ()
                       -> Handle
+                      -> (Handle -> Bool)
                       -> Severity
                       -> IO (GenericHandler Handle)
-streamHandlerWithLock customTerminalAction lock handle severity =
-    streamHandler handle customTerminalAction lock severity
+streamHandlerWithLock customTerminalAction lock handle shouldPrintError severity =
```
chshersh: `streamHandlerWithLock` function is not needed because all it does is just calls `streamHandler` with the same arguments. This function is left after my refactoring... It did more previously but now there's no actual need in this function as I see.

src/System/Wlog/Wrapper.hs
```diff
@@ -40,10 +40,11 @@ import           System.Wlog.Severity       (Severity (..))
 streamHandlerWithLock :: (Handle -> Text -> IO ())
                       -> MVar ()
                       -> Handle
+                      -> (Handle -> Bool)
```
chshersh: Could you please rename `Wrapper` module to `Terminal` module? `Wrapper` is really odd name... It's such a name only by historical reasons. Would you mind perform a little bit extra refactoring? :)

src/System/Wlog/Handler.hs
```diff
@@ -54,6 +54,9 @@ class LogHandler a where
     -- | Gets the current level.
     getLevel :: a -> Severity
 
+    -- | Needed for printing 'Error' only to @stderr@
+    shouldPrintError :: a -> Bool
```
chshersh: Approach with `shouldPrintError` is quite hacky and dirty... I think you can agree with this. And it's not quite obvious which solution to use. I have several proposals:

1. We can change `getLevel` function to return `Set Severity` for this `LogHandler`. This approach allows us to specify maximum flexibility for defining which severity should this logger use. For example, currently it's not possible to log only `Debug` messages with some logger. But with this approach it should be possible. Though, we might think more about changing configuration... It becomes really strange that our functions work with `Set` but in `yaml` files for our logger config we specify only one severity.
2. Less flexible approach is to define range of `Severity` instead of `Set`. Like `lowestSeverity :: a -> Severity` and `greatestSeverity :: a -> Severity`. This approach has less power than the previous one. But it's easier to implement and it's easier to migrate from previous config.
3. Instead of modifying `getLevel` function we can add method `excludedSeverities :: a -> Set Severity`. So it's more like your `shouldPrintError` but generalized to allow you to exclude any severity you want. This plays well with current notion of _lowestLevelOfSeverity_ and doesn't require you to change config format in backwards incompatible way.
4. Maybe something else.

What do you think @vrom911? Which one do you like more? Or what other ideas do you have?

src/System/Wlog/Handler.hs
```diff
@@ -63,10 +66,11 @@ class LogHandler a where
     -- given by the most recent call to 'setLevel'.
     handle :: a -> LogRecord -> LoggerName -> IO ()
```
chshersh: Well, this function came from `hslogger`. It's really strange that it has default implementation and it's never implemented by any logger. Could you please move this function from this type class to top level? And rename it... `handle` name is really toooooo generic and awkward. Probably something like `outputMessage` or `logHandlerMessage` might be better.

Also, one more thing I don't like in current design of `log-warper`: too many intermediate calls between something meaningful actually happens in `handle`. Let me explain:

1. Current `logMessage` is two-lines function which calls `dispatchMessage`.
```haskell
logMessage severity t = do
    name <- askLoggerName
    dispatchMessage name severity t
```
2. `dispatchMessage` for `IO` is one line function which just calls `logM`.
```haskell
dispatchMessage name prior msg = logM name prior msg
```
3. `logM` function is two-lines function which just calls `logL`:
```haskell
logM logname pri msg = do
    l <- getLogger logname
    logL l pri msg
```
4. `logL` is one line function which finally calls `handle`:
```haskell
logL l pri msg = handle l (LR pri msg) (const True)
```

I think this can be done better... We can remove `logMessage` from this chain because this function should remain polymorphic. But the others... Unless functions from this chain are not used anywhere else, we can remove intermediate redundant functions.

Also, I think function like `debugM` are not needed anymore:
https://github.com/serokell/log-warper/blob/4a0dc0922f7f487fe35f054169a56aeee923afff/src/System/Wlog/Logger.hs#L161

chshersh (changes requested):
Looks really good! I'm really glad these changes are finally happening ^.^ But more refactoring is required to make code better.

src/System/Wlog/Wrapper.hs
vrom911: Of course, it's more suitable name :slightly_smiling_face:

src/System/Wlog/Wrapper.hs
vrom911: Great! I also don't see any sense to have this one.

src/System/Wlog/Handler.hs
vrom911: I'm voting for the first one. As you said it's the most flexible and configurable for any needs :100:

src/System/Wlog/Handler.hs
vrom911: I agree, too many functions, but here I see that only `logL` could be thrown away, cause it's the only usage for the function. Also there is `logLCond`, with the same issue..
I guess others should stay :thinking:

src/System/Wlog/Handler.hs
chshersh: Okay, so let's go with 1st option :1st_place_medal:

chshersh:
@vrom911 It's getting better! Several things to refactor to improve quality of `log-warper`:

1. Could you please remove `releaseAllHandlers` from `Terminal` module and change type of `removeAllHandlers` from `IO` to `MonadIO m => m`?
2. Could you please move `setSeverity` and `setSeverity` functions from this module to `Logger` modules?
3. Rename `Handler` module to `LogHandler`.
4. Lift all functions inside `Logger` module to `MonadIO` please (and remove redundant `liftIO` from other places). And I think you can remove `trapLogging` function, it's not used anywhere.
5. Rename `Logger` module to `IOLogger`.
6. `debugM` `errorM` `infoM` `noticeM` etc. functions are not needed anymore.

examples/Playground.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE NoImplicitPrelude #-}
```
chshersh: You don't need `NoImplicitPrelude` in this file because executable target `play-log` already contains `NoImplicitPrelude` inside `default-extensions`. As I see now, `NoImplicitPrelude` is also inside `default-extensions` for `library`. So I think we can remove `{-# LANGUAGE NoImplicitPrelude #-}` from every file and nothing should break.

src/System/Wlog/CanLog.hs
```diff
@@ -31,7 +31,7 @@ import qualified Control.Monad.State.Lazy  as StateLazy
 import           Control.Monad.Trans       (MonadTrans (lift))
 
 import           System.Wlog.HasLoggerName (HasLoggerName (..))
-import           System.Wlog.Logger        (logM)
+import           System.Wlog.IOLogger      (logM)
```
chshersh: Could please perform _eta-reduction_ and change

```haskell
dispatchMessage name prior msg = logM name prior msg
```

to

```haskell
dispatchMessage = logM
```
?

src/System/Wlog/IOLogger.hs
```diff
@@ -294,10 +258,23 @@ setLevel p = lLevel .~ Just p
 clearLevel :: Logger -> Logger
 clearLevel = lLevel .~ Nothing
 
+-- | Set severity for given logger. By default parent's severity is used.
+setSeverity :: MonadIO m => LoggerName -> Severity -> m ()
+setSeverity name =
+    liftIO . updateGlobalLogger name . setLevel
```
chshersh: You don't need `liftIO` here because `updateGlobalLogger` is already lifted. And you can implement this function in one line, no need in moving function body to separate line.

src/System/Wlog/IOLogger.hs
```diff
@@ -294,10 +258,23 @@ setLevel p = lLevel .~ Just p
...
+
+-- | Set or clear severity.
+setSeverityMaybe
+    :: MonadIO m
+    => LoggerName -> Maybe Severity -> m ()
+setSeverityMaybe name Nothing =
+    liftIO $ updateGlobalLogger name clearLevel
```
chshersh: Same as well (regarding `liftIO` and alignment).

src/System/Wlog/IOLogger.hs
```diff
@@ -308,41 +285,26 @@ saveGlobalLogger l =
...
 -- >                    (setLevel DEBUG . setHandlers [s])
 updateGlobalLogger
-    :: LoggerName         -- ^ Logger name
+    :: MonadIO m
+    => LoggerName         -- ^ Logger name
     -> (Logger -> Logger) -- ^ Function to call
-    -> IO ()
+    -> m ()
 updateGlobalLogger ln func =
     do l <- getLogger ln
```
chshersh: Could you please align `do` here according to our style-guide

src/System/Wlog/IOLogger.hs
```diff
@@ -308,41 +285,26 @@ saveGlobalLogger l =
...
-        case desc of
-            "" -> ""
-            x  -> x <> ": "
-    handler :: SomeException -> IO a
-    handler e = do
-        logM logger priority (realdesc <> show e)
-        throwM e -- Re-raise it
+-- | Lifted alias to 'removeAllHandlers'.
+releaseAllHandlers :: MonadIO m => m ()
+releaseAllHandlers = removeAllHandlers
```
chshersh: As I said, `releaseAllHandlers` function is not needed anymore.

src/System/Wlog/LoggerConfig.hs
```diff
@@ -57,24 +57,25 @@ module System.Wlog.LoggerConfig
...
-import           Data.Yaml                  (FromJSON (..), ToJSON (..), Value (Object),
-                                             object, (.!=), (.:), (.:?), (.=))
-import           Formatting                 (bprint, shown)
-import           GHC.Generics               (Generic)
-import           System.FilePath            (normalise)
-
-import           System.Wlog.Handler.Simple (defaultHandleAction)
-import           System.Wlog.LoggerName     (LoggerName)
-import           System.Wlog.Severity       (Severity)
+import           Control.Lens                  (at, makeLenses, zoom, _Just)
```
chshersh: I think you can start using our new `.stylish-haskell.yaml` config (by copy-pasting it to this project directory) with our new style-guide rules for imports. You don't need to walk through each file and apply formatting rules manually. It's enough to do so only for files where you do something.

chshersh (changes requested):

src/System/Wlog/Launcher.hs
```diff
@@ -69,7 +70,7 @@ setupLogging mTimeFunction LoggerConfig{..} = do
                             customTerminalAction
                             isShowTime
                             isShowTid
-                            _lcTermSeverity
+                            (_lcTermSeverity >>= pure . severityPlus)
```
chshersh: This can be written as `severityPlus <$> _lcTermSeverity`.

src/System/Wlog/LogHandler.hs
```diff
@@ -63,8 +63,8 @@ class LogHandler a where
 
     -- | Logs an event if it meets the requirements
     -- given by the most recent call to 'setLevel'.
-    handle :: MonadIO m => a -> LogRecord -> LoggerName -> m ()
-    handle h lr@(LR pri _) logname =
+    logHandlerMessage :: MonadIO m => a -> LogRecord -> LoggerName -> m ()
+    logHandlerMessage h lr@(LR pri _) logname =
```
chshersh: This function should be moved out of this type class. It has default implementation. And nobody uses other implementation. This is very confusing. So this functions should be moved to top-level out from this type class. If somebody one day will need to override this implementation we can always move back to this type class.

src/System/Wlog/Severity.hs
```diff
@@ -26,5 +39,56 @@ data Severity
...
 
+-- | Set of 'Severity'.
+type Severities = Set Severity
+
 -- | Internal type of log records.
 data LogRecord = LR !Severity !Text deriving Show
+
+-- | 'Set' of all 'Severity's.
+allSeverities :: Set Severity
+allSeverities = Set.fromList [Debug, Info, Notice, Warning, Error]
```
chshersh: In future we might add more severities. Thus we will need to update `allSeverities` list. This is not very convenient, can be easily forgotten and compiler can't help us here. I propose to use

```haskell
allSeverities = Set.fromAscList [minBound .. maxBound]
```

This works because `Severity` is `Bounded` and `Enum`.

src/System/Wlog/Severity.hs
```diff
@@ -26,5 +39,56 @@ data Severity
...
+
+-- | Returns the 'Set' of 'Severity's of elements greater than the given value.
+severityPlus :: Severity -> Set Severity
+severityPlus s = Set.filter (>= s) allSeverities
```
chshersh: Another possible implementation of this function is to use `Enum` and `Bounded` instances for `Severity`:

```haskell
severityPlus s = Set.fromAscList [s .. maxBound]
```

src/System/Wlog/Severity.hs
```diff
@@ -26,5 +39,56 @@ data Severity
...
+type Severities = Set Severity
+
 -- | Internal type of log records.
 data LogRecord = LR !Severity !Text deriving Show
+
+-- | 'Set' of all 'Severity's.
+allSeverities :: Set Severity
+allSeverities = Set.fromList [Debug, Info, Notice, Warning, Error]
+
+-- | Returns the 'Set' of 'Severity's of elements greater than the given value.
```
chshersh: In documentation it's said that elements _are greater_ while in implementation they are _greater or equal_.

src/System/Wlog/Severity.hs
```diff
@@ -26,5 +39,56 @@ data Severity
...
+severityPlus :: Severity -> Set Severity
+severityPlus s = Set.filter (>= s) allSeverities
+
+-- | Returns 'Set' of 'Severity's not less than 'Debug'.
+debugPlus :: Set Severity
+debugPlus = allSeverities
```
chshersh: I think it's better to use `severityPlus Debug` in case we add new severity lower than `Debug`.

src/System/Wlog/Severity.hs
```diff
@@ -26,5 +39,56 @@ data Severity
...
+infoPlusWoError :: Set Severity
+infoPlusWoError = excludeError infoPlus
+
+-- | Similar to 'noticePlus' but excludes 'Error'.
+noticePlusWoError :: Set Severity
+noticePlusWoError = excludeError noticePlus
+
+-- | Similar to 'warningPlus' but excludes 'Error'.
+warningPlusWoError :: Set Severity
+warningPlusWoError = Set.singleton Warning
```
chshersh: I don't think we need too many extra functions (especially if they're not used in library itself)... Approach with `Set` is very flexible. It allows users to write their own functions to define their sets of severities. I don't think we should define every possible combination as a separate function. `debugPlus` `infoPlus` and company are very helpful! But `warningPlusWoError` is too specific. We need to exclude `Error` only in one place in our code.

src/System/Wlog/Terminal.hs
```diff
@@ -51,20 +51,20 @@ initTerminalLogging :: MonadIO m
...
-                    -> Maybe Severity
+                    -> Maybe Severities
                     -> m ()
 initTerminalLogging
     timeF
     customConsoleAction
     isShowTime
     isShowTid
-    (fromMaybe Warning -> defaultSeverity)
+    (fromMaybe (warningPlusWoError) -> defaultSeverity)
```
chshersh: Here should be only `warningPlus`. You excludes error just several lines below.

chshersh (changes requested):

chshersh (approved):

PR #53, opened 2017-11-16
[LW-30] Support stdout stderr configurations

### LINKED ISSUE #32: Change YAML config format for logger names
Currently you need to specify root logger at top level. This brings some amount of complexity to parsing `.yaml` config because you need to **NOT FORGET** to update non-logger names:

https://github.com/serokell/log-warper/blob/b219e7bd5e65b080ed59b44452c845d449200b69/src/System/Wlog/LoggerConfig.hs#L134

So logger tree should be moved into separate tree.

### LINKED ISSUE #48: [LW-30] Option to log messages only to `stderr`

### LINKED ISSUE #49: Add `WithLoggerIO` constraint
It would be very useful to have `(MonadIO m, WithLogger m)` = `WithLoggerIO m`

### LINKED ISSUE #50: Add function that will deal with `forkIO`-like functions
As it was very often pattern I've implemented 
```
forkLogIO :: (MonadIO m, WithLogger m) => LoggerNameBox IO () -> m ThreadId
forkLogIO action = do
    logName <- askLoggerName
    liftIO $ forkIO $ usingLoggerName logName $ action
```
It would be helpful to have something more common not related only to `forkIO`.

vrom911:
Resolves #48 
Resolves #32 
Resolves #49 
Resolves #50

### DIFF
--- a/src/System/Wlog/LoggerConfig.hs
+++ b/src/System/Wlog/LoggerConfig.hs
@@ -274,18 +274,6 @@ instance Monoid LoggerConfig where
         orCombiner  field = field lc1 <|> field lc2
         andCombiner field = field lc1  <> field lc2
 
-
-topLevelParams :: [Text]
-topLevelParams =
-    [ "rotation"
-    , "termSeveritiesOut"
-    , "termSeveritiesErr"
-    , "showTime"
-    , "showTid"
-    , "printOutput"
-    , "filePrefix"
-    ]
-
 instance FromJSON LoggerConfig where
     parseJSON = withObject "rotation params" $ \o -> do
         _lcRotation        <-         o .:? "rotation"
@@ -294,7 +282,7 @@ instance FromJSON LoggerConfig where
         _lcShowTime        <- Any <$> o .:? "showTime"    .!= False
         _lcShowTid         <- Any <$> o .:? "showTid"     .!= False
         _lcFilePrefix      <-         o .:? "filePrefix"
-        _lcTree            <- parseJSON $ Object $ filterObject topLevelParams o
+        _lcTree            <-         o .:? "loggerTree"  .!= mempty
 
         printConsoleFlag    <- o .:? "printOutput" .!= False
         let _lcConsoleAction = Last $ bool Nothing (Just defaultHandleAction) printConsoleFlag
--- a/src/System/Wlog/LoggerConfig.hs
+++ b/src/System/Wlog/LoggerConfig.hs
@@ -73,11 +73,12 @@ import System.FilePath (normalise)
 
 import System.Wlog.LoggerName (LoggerName)
 import System.Wlog.LogHandler.Simple (defaultHandleAction)
-import System.Wlog.Severity (Severities, Severity, allSeverities, debugPlus, errorPlus, infoPlus,
-                             noticePlus, warningPlus)
+import System.Wlog.Severity (Severities, allSeverities, debugPlus, errorPlus, infoPlus, noticePlus,
+                             warningPlus)
 
 import qualified Data.HashMap.Strict as HM hiding (HashMap)
 import qualified Data.Set as Set
+import qualified Data.Text as T
 import qualified Data.Text.Buildable as Buildable
 import qualified Data.Vector as Vector
 
@@ -87,6 +88,22 @@ import qualified Data.Vector as Vector
 
 filterObject :: [Text] -> HashMap Text a -> HashMap Text a
 filterObject excluded = HM.filterWithKey $ \k _ -> k `notElem` excluded
+
+parseSeverities :: Object -> Text -> Parser (Maybe Severities)
+parseSeverities o term = do
+    case HM.lookup term o of
+        Just value -> case value of
+            String word -> case word of
+                "All"      -> pure $ Just allSeverities
+                "Debug+"   -> pure $ Just debugPlus
+                "Info+"    -> pure $ Just infoPlus
+                "Notice+"  -> pure $ Just noticePlus
+                "Warning+" -> pure $ Just warningPlus
+                "Error+"   -> pure $ Just errorPlus
+                _          -> fail $ T.unpack $ "Unknown severity: " <> word
+            Array sevs  -> Just . Set.fromList . Vector.toList <$> Vector.mapM parseJSON sevs
+            _           -> fail "Incorrect severities format"
+        Nothing    -> pure $ Nothing
 
 ----------------------------------------------------------------------------
 -- LoggerTree
@@ -108,7 +125,7 @@ type LoggerMap = HashMap Text LoggerTree
 data LoggerTree = LoggerTree
     { _ltSubloggers :: !LoggerMap
     , _ltFiles      :: ![HandlerWrap]
-    , _ltSeverity   :: !(Maybe Severity)
+    , _ltSeverity   :: !(Maybe Severities)
     } deriving (Generic, Show)
 
 makeLenses ''LoggerTree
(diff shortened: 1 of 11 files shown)


src/System/Wlog/LoggerConfig.hs
```diff
@@ -256,12 +268,14 @@ topLevelParams =
...
     parseJSON = withObject "rotation params" $ \o -> do
-        _lcRotation      <-         o .:? "rotation"
-        _lcTermSeverity  <-         o .:? "termSeverity"
-        _lcShowTime      <- Any <$> o .:? "showTime"    .!= False
-        _lcShowTid       <- Any <$> o .:? "showTid"     .!= False
-        _lcFilePrefix    <-         o .:? "filePrefix"
-        _lcTree          <- parseJSON $ Object $ filterObject topLevelParams o
+        _lcRotation        <-                         o .:? "rotation"
+        -- TODO: Support word "All" which will include all severities at once
+        _lcTermSeverityOut <- (fmap Set.fromList) <$> o .:? "termSeveritiesOut"
```
chshersh: We have [`<<$>>`](http://hackage.haskell.org/package/universum-0.7.1.1/docs/Functor.html#v:-60--60--36--62--62-) in `unversum` to fight with nested functors.

src/System/Wlog/Severity.hs
```diff
@@ -44,6 +47,10 @@ data LogRecord = LR !Severity !Text deriving Show
 allSeverities :: Set Severity
 allSeverities = Set.fromAscList [minBound .. maxBound]
 
+-- | Empty 'Set' of 'Severity'.
+emptySeverities :: Severities
+emptySeverities = Set.empty
```
chshersh: I think we don't need `emptySeverities` function because it's just `mempty` (`Set` is a `Monoid`).

src/System/Wlog/Severity.hs
```diff
@@ -71,3 +78,6 @@ errorPlus = Set.singleton Error
 -- | Excludes 'Error' from the 'Set' of 'Severity's.
 excludeError :: Set Severity -> Set Severity
 excludeError = Set.delete Error
+
+unionSeverities :: Severities -> Severities -> Severities
+unionSeverities = Set.union
```
chshersh: I think we don't need `unionSeverities` function because `Set` is `Semigroup` and has `<>` operator [which is implemented](http://hackage.haskell.org/package/containers-0.5.10.2/docs/src/Data.Set.Internal.html#line-286) just like `union`.

src/System/Wlog/Terminal.hs
```diff
@@ -50,23 +51,31 @@ initTerminalLogging :: MonadIO m
...
+    maybeSevOut
+    maybeSevErr
   = liftIO $ do
     lock <- liftIO $ newMVar ()
+    let (severitiesOut, severitiesErr) =
+          case (maybeSevOut, maybeSevErr) of
+              (Nothing, Nothing)   -> (excludeError debugPlus, errorPlus)
+              (Just out, Nothing)  -> (out, emptySeverities)
+              (Nothing, Just err)  -> (emptySeverities, err)
+              (Just out, Just err) -> (out, err)
```
chshersh: Am I right that to redirect all output to `stderr` user needs to pass `Nothing` as `maybeSevOut` and `Just allSeverities` as `maybeSevErr`?

src/System/Wlog/LoggerConfig.hs
```diff
@@ -256,12 +268,14 @@ topLevelParams =
...
+        _lcTermSeverityErr <- (fmap Set.fromList) <$> o .:? "termSeveritiesErr"
```
chshersh: @vrom911 Wouldn't you mind to fix issue #32 as well in this PR? Please, ask questions if you need more details on this!

src/System/Wlog/LoggerConfig.hs
```diff
@@ -256,12 +268,14 @@ topLevelParams =
...
 instance FromJSON LoggerConfig where
     parseJSON = withObject "rotation params" $ \o -> do
-        _lcRotation      <-         o .:? "rotation"
-        _lcTermSeverity  <-         o .:? "termSeverity"
-        _lcShowTime      <- Any <$> o .:? "showTime"    .!= False
-        _lcShowTid       <- Any <$> o .:? "showTid"     .!= False
-        _lcFilePrefix    <-         o .:? "filePrefix"
-        _lcTree          <- parseJSON $ Object $ filterObject topLevelParams o
+        _lcRotation        <-                         o .:? "rotation"
+        -- TODO: Support word "All" which will include all severities at once
```
chshersh: Yes, this is really good idea! Unfortunately, it requires advanced knowledges of `aeson` library to implement this feature... I'm not an `aeson` expert so I can propose only this solution:

1. Perform lookup by key `"termSeveritiesOut"` in `o` (this is possible because [`o` is an `Object`, which is just `HashMap`](http://hackage.haskell.org/package/aeson-1.2.3.0/docs/Data-Aeson.html#t:Object)).
2. Pattern match on result of your lookup. If `Nothing` then `_lcTermSeveritiesOut = Nothing`.
3. If `Just value` then pattern match on the [`value` constructor](http://hackage.haskell.org/package/aeson-1.2.3.0/docs/Data-Aeson.html#t:Value). We're interested only in two constructors: `String` and `Array`.
4. If `Text` inside `String` constructor equals to `All` then return `allSeverities` (otherwise `fail`).
5. If `Array` then call [`Vector.mapM`](http://hackage.haskell.org/package/vector-0.12.0.1/docs/Data-Vector.html#v:mapM)` parseJSON` over resulting array.

I agree, this ugly... But I just don't know how to write better :(

By the way, moreover, we can go even further! Once this implemented, we can support more keywords other than `All` for convenience, like `Debug+`, `Info+`, etc. This will be really handy because often you cares about some level of `Severity` and all levels above that level.

chshersh (changes requested):
Looks like really good start! Waiting for fixes.

src/System/Wlog/LoggerConfig.hs
vrom911: Okay, this sounds pretty useful. Also `log-warper` would need good documentation to explain how to use this kind of features.

src/System/Wlog/LoggerConfig.hs
vrom911: Okay, I'll have a look at this one as well, thanks! :wink:

src/System/Wlog/Severity.hs
vrom911: @ChShersh Oh, right! No need in useless functions.

src/System/Wlog/Terminal.hs
vrom911: yes, true. Ideally that would be this line in the `.yaml`:
```
termSeveritiesErr: All
```

src/System/Wlog/LoggerConfig.hs
```diff
@@ -268,19 +270,26 @@ topLevelParams =
...
         let _lcConsoleAction = Last $ bool Nothing (Just defaultHandleAction) printConsoleFlag
         let _lcMapper        = mempty
         return LoggerConfig{..}
+          where
+            parseSeverities :: Object -> Text -> Parser (Maybe Severities)
+            parseSeverities o term = do
+                case HM.lookup term o of
+                    Just value -> case value of
+                                String "All" -> pure $ Just allSeverities
+                                Array sevs   -> Just . Set.fromList . Vector.toList <$> Vector.mapM parseJSON sevs
```
chshersh: You should add `_` case here if you don't want to see _Non-exhaustive pattern matching_ erro. I wonder why `-Wall` didn't warned you about that... Rather than that, looks good. You can add `Debug+` and etc. functions :)

chshersh:
Looks really good! Waiting for CHANGELOG to be updated :)

chshersh (approved):

PR #58, opened 2017-11-26
[WIP] [LW-26] Add literate haskell How-To example

vrom911:
This is only one part of the plan of writing good tutorial and how-to.
I used literate haskell to make how-to integrate `log-warper` in very simple example module.
I also created the executable for that, so you can check `how-to` to see it working.

### DIFF
--- a/examples/HowTo.md
+++ b/examples/HowTo.md
@@ -1,0 +1,147 @@
+# `log-warper` how-to
+
+If you want to see the working example first, please, check out this [very simple example](https://github.com/serokell/log-warper/blob/master/examples/Playground.hs).
+
+Here we will try to explain step by step how to integrate `log-warper` into your project.
+
+All code below can be compiled and run with these commands
+
+```bash
+stack build log-warper
+stack exec how-to
+```
+and you can see how it's working both with and without logging.
+
+Since this is a literate haskell file, we need to specify all our language extensions and imports up front.
+
+``` haskell
+{-# LANGUAGE OverloadedStrings #-}
+
+module Main (main) where
+
+
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import Data.Monoid ((<>))
+import Data.Text (Text)
+import System.Wlog (WithLoggerIO, buildAndSetupYamlLogging, logError,
+                    logInfo,productionB, usingLoggerName)
+
+import qualified Data.Text as T
+import qualified Data.Text.IO as TIO
+
+```
+
+## Adding a new logger
+
+First of all, you may ask what are we doing? What the hell is logger?
+
+**Logger** is the [output handler](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-LogHandler.html#t:LogHandler) labeled by the unique [logger name](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-LoggerName.html#t:LoggerName).
+
+To add a new logger for your needs you'll have to consider several things.
+Firstly, you should consider choosing some reasonable name for the logger you creating according to the actions that it would cover with logging. From now on let's assume that you need to have `new-logger` in your project.
+
+### Configuration file
+
+To configure your logger you will need configuration `.yaml` file.
+In our example we will use very simple one with the minimal settings needed for the file to work:
+
+```yaml
+loggerTree:
+    severity: Warning+    # severities for «root» logger
+    new-logger:           # logger named «new-logger»
+        severity: Debug+  # severities for logger «new-logger»
+```
+
+For full information about all configuration file's features and options please see [this section](#link-to-readme-config).
+
+### Setting up
+
+Before using your logger you need to set it up. To do that there are very useful functions in `log-warper`.
+In our example we will run all functions with our logger from `main` function of `Main` module.
+
+Let's consider that this is what we have:
+
+```haskell
+mainWithoutLogging :: IO ()
+mainWithoutLogging = inputLength
+```
+ where `inputLength` is the function where we will add some logging messages.
+
+So to make it possible we should transformed this function to the following:
+
+```haskell
+mainWithLogging :: IO ()
+mainWithLogging = do
+    buildAndSetupYamlLogging productionB "examples/how-to-log-config.yaml"
+    usingLoggerName "new-logger" inputLengthWithLog
+```
+where we set up the config from the file and run `inputLengthWithLog` under the logger with name `new-logger`.
+
+Let's have a closer look at how we transform the `inputLength` function to the `inputLengthWithLog` that would work with logging in the next paragraph.
+
+## Add logging messages to function in `IO`
+
+If you want to use logging in a function that already runs in the `IO` monad, replace the monad with [`WithLoggerIO`](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-CanLog.html#t:WithLoggerIO) and `liftIO` all your `IO` actions.
+
+For example, our `inputLength` function works this way:
+```haskell
+
+inputLength :: IO ()
+inputLength = do
+    input <- TIO.getLine
+    case T.length input of
+        0 -> putStrLn "Should not be empty" >> inputLength
+          -- v ^  you need to log these
+        _ -> putStrLn "Well done" >> funWithInput input
+
+```
+Very easy, nothing interesting. We have some input from console and write some messages depending of the length of user's input.
+What we are going to do is to log this messages instead of just `putStrLn` them.
+
+First that we should consider is that logging can not be done in `IO`, so we need make the function work in `WithLoggerIO` monad,
+so that we are able to do any logging stuff inside.
+After this is done you can easily use `logDebug`, `logInfo`, `logWarning` etc.
+(see others [here](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-CanLog.html#v:logDebug)) to log messages.
+As it's not `IO` after our transformations, we shouldn't forget to `liftIO` functions that work in `IO` and doesn't have logging inside.
+As result we will get
+
+```haskell
+
+inputLengthWithLog :: WithLoggerIO m => m ()
+inputLengthWithLog = do
+    input <- liftIO TIO.getLine
+    case T.length input of
+        0 -> do
+            logError "Should not be empty"
+            inputLengthWithLog
+        _ -> do
+            logInfo "Well done"
+            funWithInputInMonad input
+
+```
+
+Another moment, let's have a closer look at the `funWithInput` function from the example above.
+We see that its type is `IO ()` and it doesn't have logging inside.
+
+```haskell
+funWithInput :: Text -> IO ()
+funWithInput input = do
+    TIO.putStrLn $ "Wow! You wrote " <> input
+
+```
+So we also can rewrite it in this way
+```haskell
+funWithInputInMonad :: MonadIO m => Text -> m ()
+funWithInputInMonad input =
+    liftIO $ TIO.putStrLn $ "Wow! You wrote " <> input
+
+```
+or usage of `liftIO` is also the solution.
+
+And as a conclusion, let's write the `main` function to check how we managed to apply logging in very simple example
+```haskell
+main :: IO ()
+main = do
+    mainWithoutLogging
+    mainWithLogging
+```
--- a/examples/HowTo.md
+++ b/examples/HowTo.md
@@ -1,8 +1,10 @@
 # `log-warper` how-to
 
-If you want to see the working example first, please, check out this [very simple example](https://github.com/serokell/log-warper/blob/master/examples/Playground.hs).
+If you want to see the working example first, please, check out this
+[very simple example](https://github.com/serokell/log-warper/blob/master/examples/Playground.hs).
 
-Here we will try to explain step by step how to integrate `log-warper` into your project.
+Here we will try to explain step by step how to integrate `log-warper` into our very small project,
+specifically how to replace `putStrLn` with simple naive logging.
 
 All code below can be compiled and run with these commands
 
@@ -12,7 +14,10 @@ stack exec how-to
 ```
 and you can see how it's working both with and without logging.
 
-Since this is a literate haskell file, we need to specify all our language extensions and imports up front.
+## Preamble: imports and language extensions
+
+Since this is a literate haskell file, we need to specify all our language extensions
+and imports up front.
 
 ``` haskell
 {-# LANGUAGE OverloadedStrings #-}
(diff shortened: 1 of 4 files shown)


examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
+# `log-warper` how-to
+
+If you want to see the working example first, please, check out this [very simple example](https://github.com/serokell/log-warper/blob/master/examples/Playground.hs).
+
+Here we will try to explain step by step how to integrate `log-warper` into your project.
```
chshersh: I would elaborate here and clarify that we will tell how to replace simple naive logging with `putStrLn` using `log-warper`.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+All code below can be compiled and run with these commands
+
+```bash
+stack build log-warper
+stack exec how-to
+```
+and you can see how it's working both with and without logging.
+
+Since this is a literate haskell file, we need to specify all our language extensions and imports up front.
```
chshersh: I think it's better to bring more structure into this file and put this line with imports and pragmas into separate section like

```
## Preamble: imports and language extensions
```

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+``` haskell
+{-# LANGUAGE OverloadedStrings #-}
+
+module Main (main) where
+
+
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import Data.Monoid ((<>))
```
chshersh: I would like to import `(<>)` from `Data.Semigroup` module :) This helps more to adopt `Semigroup`.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+import Data.Text (Text)
+import System.Wlog (WithLoggerIO, buildAndSetupYamlLogging, logError,
+                    logInfo,productionB, usingLoggerName)
```
chshersh: Space after `logInfo,`

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+import qualified Data.Text as T
+import qualified Data.Text.IO as TIO
+
+```
+
+## Adding a new logger
+
+First of all, you may ask what are we doing? What the hell is logger?
+
+**Logger** is the [output handler](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-LogHandler.html#t:LogHandler) labeled by the unique [logger name](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-LoggerName.html#t:LoggerName).
```
chshersh: I think it worth adding something like this as well: 

When you log new messages you need to specify logger which is responsible for logging this message. Fortunately, `log-warper` has monadic interfaces and logger will be taken explicitly from current context if possible, thus you need to specify logger only once for some block of actions. Don't worry, this will be covered later!

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+To add a new logger for your needs you'll have to consider several things.
+Firstly, you should consider choosing some reasonable name for the logger you creating according to the actions that it would cover with logging. From now on let's assume that you need to have `new-logger` in your project.
```
chshersh: > that you need to have `new-logger` in your project

Maybe better

> that you need to have logger with name `new-logger` in your project

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Configuration file
+
+To configure your logger you will need configuration `.yaml` file.
```
chshersh: It worth adding that you also can configure everything directly in code but this is less convenient.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+In our example we will use very simple one with the minimal settings needed for the file to work:
+
+```yaml
```
chshersh: I would like to have link to existing `yaml` file in this repository somewhere near here.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+    severity: Warning+    # severities for «root» logger
+    new-logger:           # logger named «new-logger»
+        severity: Debug+  # severities for logger «new-logger»
+```
+
+For full information about all configuration file's features and options please see [this section](#link-to-readme-config).
+
+### Setting up
+
+Before using your logger you need to set it up. To do that there are very useful functions in `log-warper`.
```
chshersh: Not `using your logger` but better to say `using logging system`. Because this action configures not only single logger, but whole logging system.

chshersh: I think it's better to omit word `useful` here. Functions are not very useful, when you are ought to use them... And then not even convenient enough currently :(

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+Let's consider that this is what we have:
+
+```haskell
+mainWithoutLogging :: IO ()
+mainWithoutLogging = inputLength
+```
+ where `inputLength` is the function where we will add some logging messages.
+
+So to make it possible we should transformed this function to the following:
```
chshersh: s/transformed/transform

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+```haskell
+mainWithLogging :: IO ()
+mainWithLogging = do
+    buildAndSetupYamlLogging productionB "examples/how-to-log-config.yaml"
+    usingLoggerName "new-logger" inputLengthWithLog
+```
+where we set up the config from the file and run `inputLengthWithLog` under the logger with name `new-logger`.
```
chshersh, line 97: I think it should be added that here we use `inputLengthWithLog` because we can't just add logging automatically to existing function which works under `IO`, we need to change our faction. And `inputLengthWithLog` is a version of `inputLength` with logging added. You're already saying this, but maybe you need to slightly rearrange and move sentences. Or even repeat, but earlier. There's absolutely no problem in repeating some complex stuff! Use so called _semantic structure_ for tutorial. Here is what _semantic structure_ in Haskell:

```haskell
type Color
type Location
μ :: Image a → (Location → a)
type Region = Image Bool
type BMP = Image Color

-- along with a tight set of fn's for image manipulation
lift :: a → Image a
lift1 :: (a → b) → (Image a → Image b)
...
cond :: Region → Image a → Image a → Image a
...
````

File: examples/HowTo.md [hunk 12]
```diff
@@ -0,0 +1,147 @@
...
+
+Let's have a closer look at how we transform the `inputLength` function to the `inputLengthWithLog` that would work with logging in the next paragraph.
+
+## Add logging messages to function in `IO`
+
+If you want to use logging in a function that already runs in the `IO` monad, replace the monad with [`WithLoggerIO`](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-CanLog.html#t:WithLoggerIO) and `liftIO` all your `IO` actions.
```
chshersh: > runs in the `IO` monad, replace the monad with [`WithLoggerIO`]

Better to say like this:

> runs in the `IO`, replace concrete `IO` type with polymorphic constraint [`WithLoggerIO m => m`]

chshersh: > and `liftIO` all your `IO` actions.

Better to

> and add [`liftIO`](link to `liftIO`) all your `IO` actions.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+inputLength :: IO ()
+inputLength = do
+    input <- TIO.getLine
+    case T.length input of
+        0 -> putStrLn "Should not be empty" >> inputLength
+          -- v ^  you need to log these
+        _ -> putStrLn "Well done" >> funWithInput input
+
+```
```
chshersh: You need to add some **_Note:_** and say that all logging functions take [`Text`](link to Text) messages instead of `String`. So users might convert their `String`s to `Text`s with `T.pack` or migrate to `text` package or use some custom prelude which makes it more convenient to work with `Text`.. ;)

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+
+import Control.Monad.IO.Class (MonadIO, liftIO)
+import Data.Monoid ((<>))
+import Data.Text (Text)
+import System.Wlog (WithLoggerIO, buildAndSetupYamlLogging, logError,
+                    logInfo,productionB, usingLoggerName)
+
+import qualified Data.Text as T
+import qualified Data.Text.IO as TIO
```
chshersh: I think qualified names `Text` and `TextIO` are better than `T` and `TIO`. It makes code easier to understand. Because it's easier to remember and recognize `TextIO.getLine` rather than `TIO.getLine`.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+inputLength = do
+    input <- TIO.getLine
+    case T.length input of
+        0 -> putStrLn "Should not be empty" >> inputLength
+          -- v ^  you need to log these
+        _ -> putStrLn "Well done" >> funWithInput input
+
+```
+Very easy, nothing interesting. We have some input from console and write some messages depending of the length of user's input.
+What we are going to do is to log this messages instead of just `putStrLn` them.
```
chshersh: s/this/these

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+
+First that we should consider is that logging can not be done in `IO`, so we need make the function work in `WithLoggerIO` monad,
```
chshersh: `WithLoggerIO` is not a monad, because monad is some specific type. It's better to say that function should work in some polymorphic monad, which implements `WithLoggerIO` constraint.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+If you want to use logging in a function that already runs in the `IO` monad, replace the monad with [`WithLoggerIO`](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-CanLog.html#t:WithLoggerIO) and `liftIO` all your `IO` actions.
+
+For example, our `inputLength` function works this way:
+```haskell
+
+inputLength :: IO ()
+inputLength = do
+    input <- TIO.getLine
+    case T.length input of
+        0 -> putStrLn "Should not be empty" >> inputLength
```
chshersh: Could you please use `do` here as well? So transformation become more clear.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+```
+or usage of `liftIO` is also the solution.
+
+And as a conclusion, let's write the `main` function to check how we managed to apply logging in very simple example
+```haskell
+main :: IO ()
+main = do
+    mainWithoutLogging
+    mainWithLogging
+```
```
chshersh, line 187: I have very cool idea: at the end of tutorial you can add two-cell table (like in this project: https://github.com/serokell/importify/blob/master/README.md) with IO on the left and `WithLoggerIO` on the right. The only difficulty is whether it will affect compilation of this file... Maybe `markdown-unlit` has some special pragma with which you can disable compilation of some part of Markdown, or maybe it doesn't compile tables automatically, or maybe you will need to put this table into separate file...

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+As result we will get
+
+```haskell
+
+inputLengthWithLog :: WithLoggerIO m => m ()
+inputLengthWithLog = do
+    input <- liftIO TIO.getLine
+    case T.length input of
+        0 -> do
+            logError "Should not be empty"
```
chshersh: You can comment here that `putStrLn` here is replaced with `logError`.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+So we also can rewrite it in this way
+```haskell
+funWithInputInMonad :: MonadIO m => Text -> m ()
+funWithInputInMonad input =
+    liftIO $ TIO.putStrLn $ "Wow! You wrote " <> input
+
+```
+or usage of `liftIO` is also the solution.
+
+And as a conclusion, let's write the `main` function to check how we managed to apply logging in very simple example
```
chshersh, line 181: I still want to see screenshot with the result of running `how-to` and observing results... Like in this way:

1. `stack exec how-to`
2. Input empty string
3. Input "abacaba"
4. Input empty string.
5. Input "abacaba"
6. Make screenshot

So user can see how messages are printed.

examples/HowTo.md
```diff
@@ -0,0 +1,147 @@
...
+In our example we will use very simple one with the minimal settings needed for the file to work:
+
+```yaml
+loggerTree:
+    severity: Warning+    # severities for «root» logger
+    new-logger:           # logger named «new-logger»
+        severity: Debug+  # severities for logger «new-logger»
+```
+
+For full information about all configuration file's features and options please see [this section](#link-to-readme-config).
```
chshersh: I think that you also should add that user don't need to worry about `root` logger for now. This part is explained in section devoted to hierarchical logging.

chshersh (changes requested):

examples/HowTo.md
vrom911: Yes, there is `ignore` mark specially for these cases when you don't won't something to be compiled, so it would be cool to have the table! :heart_eyes:

vrom911: Well, the thing is that it's not possible to put code blocks into markdown tables (or I missing something..) :disappointed:

examples/HowTo.md
```diff
@@ -21,42 +26,56 @@ module Main (main) where
...
 
-To configure your logger you will need configuration `.yaml` file.
-In our example we will use very simple one with the minimal settings needed for the file to work:
+There are two ways of configuring the logger you creating. One option
+is to configure everything directly in the code, which will make your
+code less convenient. Here we will use another configuration possibility —
+we will write configurations in special `.yaml` file.
+In our example we will use very simple one with the minimal settings needed for the file to work.
+You can have a look at [`how-to-log-config.yaml`](https://github.com/serokell/log-warper/blob/master/examples/how-to-log-config.yaml),
+it contain the following lines:
```
chshersh: s/contain/contains

examples/HowTo.md
```diff
@@ -21,42 +26,56 @@ module Main (main) where
...
+according to the actions that it would cover with logging. From now on let's assume
+that you need to have the logger with the name `new-logger` in your project.
 
 ### Configuration file
 
-To configure your logger you will need configuration `.yaml` file.
-In our example we will use very simple one with the minimal settings needed for the file to work:
+There are two ways of configuring the logger you creating. One option
+is to configure everything directly in the code, which will make your
+code less convenient. Here we will use another configuration possibility —
```
chshersh: Well, code can't be less or more convenient. It's the process of writing code which can be less or more convenient..

examples/HowTo.md
```diff
@@ -76,29 +95,48 @@ mainWithLogging = do
...
-Let's have a closer look at how we transform the `inputLength` function to the `inputLengthWithLog` that would work with logging in the next paragraph.
+Let's have a closer look at how we transform the `inputLength` function
+to the `inputLengthWithLog` that would work with logging in the next paragraph.
 
 ## Add logging messages to function in `IO`
 
-If you want to use logging in a function that already runs in the `IO` monad, replace the monad with [`WithLoggerIO`](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-CanLog.html#t:WithLoggerIO) and `liftIO` all your `IO` actions.
+If you want to use logging in a function that already runs in the `IO`,
+replace concrete `IO` type with polymorphic constraint
+[`WithLoggerIO m => m`](https://hackage.haskell.org/package/log-warper-1.7.1/docs/System-Wlog-CanLog.html#t:WithLoggerIO)
```
chshersh: You can write this in a cool way to add hyperlink only to `WithLogger`: [`WithLoggerIO`](some text)` m => m`

examples/HowTo.md
chshersh: Yes, this is really a problem, indeed... I think you can use inline code blocks and find a way to use space characters there. I think that'll do for now.

chshersh (approved):

PR #59, opened 2017-11-27
[LW-32] Add Exception module (#57)

### LINKED ISSUE #57: Add 'logException' function and utilities
It would be really good to have some utility function to log exceptions with `Error` severity, something like that:

```haskell
logException :: (WithLogger m, Exception e) => e -> m ()
```

I would also appreciate to have function like `catch` which calls `logException` of some `IO` action if this exception is thrown.

vrom911:
Resolves #57

### DIFF
--- a/src/System/Wlog/Exception.hs
+++ b/src/System/Wlog/Exception.hs
@@ -1,0 +1,27 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ExplicitForAll      #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | This module deals with Exception logging.
+
+module System.Wlog.Exception
+       ( logException
+       , catchLog
+       ) where
+
+import Universum
+
+import Control.Exception.Base (Exception)
+
+import System.Wlog.CanLog (WithLogger, WithLoggerIO, logError)
+
+-- | Logs exception's description with ''System.Wlog.Severity.Error' 'System.Wlog.Severity.Severity'
+logException :: (WithLogger m, Exception e) => e -> m ()
+logException e = logError $ show e
+
+-- | Runs the action, if an exception is raised the 'logException' is executed.
+catchLog :: forall m e . (WithLoggerIO m, MonadCatch m, Exception e) => m () -> m ()
+catchLog a = a `catch` logE
+  where
+    logE :: e -> m ()
+    logE = logException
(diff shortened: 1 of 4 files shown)


src/System/Wlog/Exception.hs
```diff
@@ -0,0 +1,27 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
chshersh, line 1: Can you research and say why do you need `AllowAmbiguousTypes` in this module?

src/System/Wlog/Exception.hs
```diff
@@ -0,0 +1,27 @@
...
+       ) where
+
+import Universum
+
+import Control.Exception.Base (Exception)
+
+import System.Wlog.CanLog (WithLogger, WithLoggerIO, logError)
+
+-- | Logs exception's description with ''System.Wlog.Severity.Error' 'System.Wlog.Severity.Severity'
+logException :: (WithLogger m, Exception e) => e -> m ()
```
chshersh: I think, you need to specify `e` as a first type variable with `forall`. I expect that in future users might want to use `-XTypeApplications` to specify exception to log.

src/System/Wlog/Exception.hs
```diff
@@ -0,0 +1,27 @@
...
+logException e = logError $ show e
```
chshersh: `logException = logError . show` :)

src/System/Wlog/Exception.hs
```diff
@@ -0,0 +1,27 @@
...
+
+-- | Runs the action, if an exception is raised the 'logException' is executed.
+catchLog :: forall m e . (WithLoggerIO m, MonadCatch m, Exception e) => m () -> m ()
```
chshersh: Same here: swap `m` and `e`.

src/System/Wlog/Exception.hs
chshersh: Nevermind, I got it :slightly_smiling_face:

chshersh (changes requested):

chshersh:
Also, I would to see examples of usages of these functions. It might be not obvious how to use then, because you need to specify type of exception using `-XTypeApplications`. Adding examples to documentation with `>>>` is a good idea!

src/System/Wlog/Exception.hs
vrom911: Yes, true! This definitely should be done in right order..

src/System/Wlog/Exception.hs
```diff
@@ -0,0 +1,52 @@
...
+runWithExceptionLog = catchLog @IOException (liftIO simpleIOfun)
+
+simpleIOfun :: IO ()
+simpleIOfun = getLine >>= readFile >>= putStrLn
+
+@
+
+and when run you will get:
+
+>>> stack exec except-log
```
chshersh: You're running this example with `stack exec except-log` command, but there's no `except-log` target in this project. You can either add this target to project or use some abstract generic command in CLI like `run-main-from-this-example`.

chshersh (changes requested):

chshersh (approved):

PR #63, opened 2017-12-01
Round picoseconds by power of ten

avieth:
Previously you could round the seconds by arbitrary divisor.
Perhaps my imagination fails me but I don't think this is a useful
feature. I've come across logs in which all timestamps are rounded to 5
seconds, severly reducing their usefulness.

This patch changes the meaning of the `Int` parameter to the rounded
logging formatter. It now gives the number of least-significant digits
of the picosecond timestamps that should be forgotten, i.e. your
timestamp will be  s * 10^n  picoseconds.

### DIFF
--- a/src/System/Wlog/Formatter.hs
+++ b/src/System/Wlog/Formatter.hs
@@ -33,7 +33,7 @@ import Control.Concurrent (myThreadId)
 import Data.Monoid (mconcat)
 import Data.Text.Lazy.Builder as B
 import Data.Time (formatTime, getCurrentTime, getZonedTime)
-import Data.Time.Clock (UTCTime (..))
+import Data.Time.Clock (UTCTime (..), diffTimeToPicoseconds, picosecondsToDiffTime)
 import Data.Time.Format (FormatTime)
 import Fmt (fmt, padRightF, (+|), (|+), (|++|))
 import Fmt.Time (dateDashF, hmsF, subsecondF, tzNameF)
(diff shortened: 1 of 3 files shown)


src/System/Wlog/Formatter.hs
```diff
@@ -169,23 +169,28 @@ centiUtcTimeF t =
   where
     centiSecondF = padRightF 3 '0' . T.take 3 . fmt . subsecondF
 
+-- | Get the current time rounded to 10^n picoseconds.
```
chshersh, line 172: I think you also need to change documentation for `HandlerWrap` data type:

https://github.com/serokell/log-warper/blob/7bc6f5446fd411b457864455ab53be3a77a4ae19/src/System/Wlog/LoggerConfig.hs#L117

chshersh (changes requested):
Completely agree. Previous behavior is really strange. Moreover, you can't even specify more precise rounding with previous setting :( Only less precise 😕 

I guess this function from SO answer is supposed to be used instead of current implementation of `roundBy` but something went wrong...

https://stackoverflow.com/a/12450771/2900502

Also, could you please update CHANGES.md? Version `1.7.2` is not on Hackage yet. So you can freely describe your change there!
https://github.com/serokell/log-warper/blob/master/CHANGES.md

src/System/Wlog/Formatter.hs
avieth: Done

chshersh (approved):

volhovm:
Nice. Somebody should integrate it into CSL though.

chshersh:
@volhovm I've uploaded version 1.7.2 to Hackage, so integration is possible.

PR #68, opened 2017-12-06
[#61] Add functions to simplify the run with logging

### LINKED ISSUE #61: Make log running interface simpler
It would be good to have simpler interface for running logging. For example, this default running from tutorial can be done in separate function like `runLoggingWithFile`:

```haskell
mainWithLogging :: IO ()
mainWithLogging = do
    buildAndSetupYamlLogging productionB "examples/how-to-log-config.yaml"
    usingLoggerName "new-logger" inputLengthWithLog
```

Another function which can be helpful is to not use config from file. Just hardcode some default reasonable config (well, this default config still can be at least a function from `LoggerName` to `LoggerConfig` and emulate one from tutorial).

vrom911:
Resolves #61

### DIFF
--- a/examples/HowTo.md
+++ b/examples/HowTo.md
@@ -28,8 +28,7 @@ module Main (main) where
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Semigroup ((<>))
 import Data.Text (Text)
-import System.Wlog (WithLoggerIO, buildAndSetupYamlLogging, logError,
-                    logInfo, productionB, usingLoggerName)
+import System.Wlog (WithLoggerIO, runLoggingWithFile, logError, logInfo)
 
 import qualified Data.Text as Text
 import qualified Data.Text.IO as TextIO
@@ -90,9 +89,8 @@ So to make it possible we should transform this function to the following:
 
 ```haskell
 mainWithLogging :: IO ()
-mainWithLogging = do
-    buildAndSetupYamlLogging productionB "examples/how-to-log-config.yaml"
-    usingLoggerName "new-logger" inputLengthWithLog
+mainWithLogging =
+    runLoggingWithFile "examples/how-to-log-config.yaml" "new-logger" inputLengthWithLog
 ```
 where we set up the config from the file and run `inputLengthWithLog` under the logger with name `new-logger`.
 Note that we use `inputLengthWithLog` now instead of `inputLength` because we can not just add logging
--- a/examples/HowTo.md
+++ b/examples/HowTo.md
@@ -28,7 +28,7 @@ module Main (main) where
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Semigroup ((<>))
 import Data.Text (Text)
-import System.Wlog (WithLoggerIO, runLoggingWithFile, logError, logInfo)
+import System.Wlog (WithLoggerIO, launchFromFile, logError, logInfo)
 
 import qualified Data.Text as Text
 import qualified Data.Text.IO as TextIO
@@ -90,7 +90,7 @@ So to make it possible we should transform this function to the following:
 ```haskell
 mainWithLogging :: IO ()
 mainWithLogging =
-    runLoggingWithFile "examples/how-to-log-config.yaml" "new-logger" inputLengthWithLog
+    launchFromFile "examples/how-to-log-config.yaml" "new-logger" inputLengthWithLog
 ```
 where we set up the config from the file and run `inputLengthWithLog` under the logger with name `new-logger`.
 Note that we use `inputLengthWithLog` now instead of `inputLength` because we can not just add logging
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -8,8 +8,8 @@ import Data.Monoid ((<>))
 import Data.Yaml.Pretty (defConfig, encodePretty)
 
 import System.Wlog (CanLog, buildAndSetupYamlLogging, logDebug, logError, logInfo, logNotice,
-                    logWarning, modifyLoggerName, parseLoggerConfig, prefixB, productionB,
-                    removeAllHandlers, usingLoggerName)
+                    logWarning, modifyLoggerName, parseLoggerConfig, productionB, removeAllHandlers,
+                    usingLoggerName)
 
 testLoggerConfigPath :: FilePath
 testLoggerConfigPath = "logger-config-example.yaml"
@@ -17,7 +17,7 @@ testLoggerConfigPath = "logger-config-example.yaml"
 testToJsonConfigOutput :: MonadIO m => m ()
 testToJsonConfigOutput = do
     cfg             <- parseLoggerConfig testLoggerConfigPath
-    let builtConfig  = cfg <> productionB <> prefixB "logs"
+    let builtConfig  = cfg <> productionB
     putStrLn $ encodePretty defConfig builtConfig
 
 testLogging :: (CanLog m) => m ()
@@ -43,8 +43,7 @@ showSomeLog = do
 main :: IO ()
 main = do
     testToJsonConfigOutput
-    let config = (productionB <> prefixB "logs")
     bracket_
-        (buildAndSetupYamlLogging config testLoggerConfigPath)
+        (buildAndSetupYamlLogging productionB testLoggerConfigPath)
         removeAllHandlers
         (testLogging >> showSomeLog)
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -7,9 +7,9 @@ import Universum
 import Data.Monoid ((<>))
 import Data.Yaml.Pretty (defConfig, encodePretty)
 
-import System.Wlog (CanLog, buildAndSetupYamlLogging, logDebug, logError, logInfo, logNotice,
-                    logWarning, modifyLoggerName, parseLoggerConfig, productionB, removeAllHandlers,
-                    usingLoggerName)
+import System.Wlog (CanLog, defaultConfig, launchFromFile, launchSimpleLogging, logDebug, logError,
+                    logInfo, logNotice, logWarning, modifyLoggerName, parseLoggerConfig,
+                    productionB, usingLoggerName)
 
 testLoggerConfigPath :: FilePath
 testLoggerConfigPath = "logger-config-example.yaml"
@@ -33,7 +33,7 @@ testLogging = usingLoggerName "node" $ do
 
     logError   "BARDAQ"
 
-showSomeLog :: IO ()
+showSomeLog :: (CanLog m, MonadIO m) => m ()
 showSomeLog = do
     putText "Other log:"
     usingLoggerName "naked" $ do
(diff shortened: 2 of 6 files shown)


examples/HowTo.md
```diff
@@ -90,9 +89,8 @@ So to make it possible we should transform this function to the following:
 
 ```haskell
 mainWithLogging :: IO ()
-mainWithLogging = do
-    buildAndSetupYamlLogging productionB "examples/how-to-log-config.yaml"
-    usingLoggerName "new-logger" inputLengthWithLog
+mainWithLogging =
+    runLoggingWithFile "examples/how-to-log-config.yaml" "new-logger" inputLengthWithLog
```
chshersh: It always makes me feel uncomfortable when function takes two arguments with the same structure near each other. Here user easily can call this function as

```haskell
runLoggingWithFile "new-logger" "examples/how-to-log-config.yaml" inputLengthWithLog
```

And he won't notice error until he actually run the program. For me it's always preferable to avoid such errors at compile time. This can be done using `newtype`s. But `LoggerName` is already `newtype`! So the problem in the fact that `LoggerName` has `IsString` instance. And we can't remove this instance because it will make things less convenient. So it's not trivial how to solve this problem at compile time.

Though, one might say, that this is not really a problem, because if you're running logging at the start of your program, you will see error immediately at runtime (unless you have `.yaml` file with the same name as `LoggerName`).

What do you think here, @vrom911?

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
 -- See this module description.
 initLoggingFromYaml :: MonadIO m => FilePath -> m ()
 initLoggingFromYaml = buildAndSetupYamlLogging mempty
+
+-- | Initializes logging using given 'FilePath' to logger configurations,
+-- runs the action with the given 'LoggerName'.
+runLoggingWithFile :: MonadIO m => FilePath -> LoggerName -> LoggerNameBox m a -> m a
```
chshersh: Maybe rename this to `launchFromFile`?

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
...
+runLoggingWithFile filename logName f = do
```
chshersh: In general, it's better to call monadic arguments as `action` variable, and `f` is for functions (and `v` is for _vendetta_).
`LoggerNameBox m a` is closer to actions rather than to functions.

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
 -- See this module description.
 initLoggingFromYaml :: MonadIO m => FilePath -> m ()
```
chshersh: I think we don't need `initLoggingFromYaml` function anymore.

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
...
+                    }
+              )
+            ]
+        , _ltFiles      = []
+        , _ltSeverity   = Just warningPlus   -- root severities are Warning+
+        }
+    }
+
+-- | Set ups the logging with default configurations and runs the action with the given 'LoggerName'.
+runDefaultLogging :: MonadIO m => LoggerName -> LoggerNameBox m a -> m a
```
chshersh: Maybe better to name it as `launchSimpleLogging`?

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
...
+
+-- | Initializes logging using given 'FilePath' to logger configurations,
+-- runs the action with the given 'LoggerName'.
+runLoggingWithFile :: MonadIO m => FilePath -> LoggerName -> LoggerNameBox m a -> m a
+runLoggingWithFile filename logName f = do
+    buildAndSetupYamlLogging productionB filename
+    usingLoggerName logName f
+
+-- | Default logging configuration with the "node" logger.
+defaultConfig :: LoggerConfig
```
chshersh: Could you please create `defaultConfig` using lenses? Unfortunately, this feature is not specified anywhere, so there was no chance for you to know about it :( But creating `LoggerConfig`s manually is the thing that supposed to be done quite often. So I invented convenient DSL for such logging initialization.

You can see this feature here:
1. https://github.com/serokell/log-warper/blob/5783fde3f8cca225e146d08173584dacfe67ed3b/test/Test/Wlog/RollingSpec.hs#L66
2. https://github.com/input-output-hk/cardano-sl/blob/8507d03ba928e07daea57f9a52f6dc03a9d65779/lib/src/Pos/Client/CLI/Util.hs#L61
3. https://issues.serokell.io/issue/LW-12

If you have any questions, don't be shy and ask them :)

src/System/Wlog/Launcher.hs
```diff
@@ -28,6 +28,8 @@ module System.Wlog.Launcher
        ( buildAndSetupYamlLogging
        , initLoggingFromYaml
        , parseLoggerConfig
+       , runLoggingWithFile
```
chshersh: Could you also change `Playground.hs` to use this function?

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
...
+    , _lcTermSeverityOut = Nothing    -- default will be used
+    , _lcTermSeverityErr = Nothing    -- default will be used
+    , _lcShowTime        = Any True
+    , _lcShowTid         = Any False
+    , _lcConsoleAction   = mempty
+    , _lcMapper          = mempty
+    , _lcFilePrefix      = Nothing
+    , _lcTree            = LoggerTree
+        { _ltSubloggers = HM.fromList     -- logger "node" with severities Debug+
+            [ ( "node"
```
chshersh: `defaultConfig` should be function from `LoggerName` to `LoggerConfig` and here you should use given `LoggerName` instead of `"node"`.

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
...
+                    , _ltSeverity   = Just debugPlus
+                    }
+              )
+            ]
+        , _ltFiles      = []
+        , _ltSeverity   = Just warningPlus   -- root severities are Warning+
+        }
+    }
+
+-- | Set ups the logging with default configurations and runs the action with the given 'LoggerName'.
```
chshersh: Could you please add into documentation link to `'defaultConfig'`?

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
...
 initLoggingFromYaml = buildAndSetupYamlLogging mempty
+
+-- | Initializes logging using given 'FilePath' to logger configurations,
+-- runs the action with the given 'LoggerName'.
+runLoggingWithFile :: MonadIO m => FilePath -> LoggerName -> LoggerNameBox m a -> m a
+runLoggingWithFile filename logName f = do
+    buildAndSetupYamlLogging productionB filename
+    usingLoggerName logName f
+
+-- | Default logging configuration with the "node" logger.
```
chshersh: I would like to see in documentation what is enabled by default in default logging. Like, time is shown, severity for root logger, severity for given logger, etc. You don't need to describe things like _logger rotation is disabled_. Describe only differences from `mempty :: LoggerConfig`.

src/System/Wlog/Launcher.hs
```diff
@@ -124,3 +127,42 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
...
+defaultConfig :: LoggerConfig
+defaultConfig = LoggerConfig
+    { _lcRotation        = Nothing
+    , _lcTermSeverityOut = Nothing    -- default will be used
+    , _lcTermSeverityErr = Nothing    -- default will be used
+    , _lcShowTime        = Any True
+    , _lcShowTid         = Any False
+    , _lcConsoleAction   = mempty
```
chshersh: You should use [`defaultHandleAction`](https://github.com/serokell/log-warper/blob/5783fde3f8cca225e146d08173584dacfe67ed3b/src/System/Wlog/LogHandler/Simple.hs#L67) here instead of `mempty` (just like it's done in `productionB`). Otherwise you won't see output to terminal

src/System/Wlog/Launcher.hs
```diff
@@ -28,6 +28,8 @@ module System.Wlog.Launcher
        ( buildAndSetupYamlLogging
        , initLoggingFromYaml
        , parseLoggerConfig
+       , runLoggingWithFile
+       , runDefaultLogging
```
chshersh: I think you should export `defaultConfig`. In case people want to extend it.

chshersh (changes requested):
Looks good! Though some code refactoring is required.

examples/HowTo.md
vrom911: @ChShersh , good point. But I don't see the way to solve it.. 
> you will see error immediately at runtime (unless you have .yaml file with the same name as LoggerName)

I think this is fair one. And also it's not an issue at all if you're naming the file and logger with the same name, two equal arguments won't let you to mess up with them :smile:

examples/Playground.hs
```diff
@@ -43,8 +43,7 @@ showSomeLog = do
 main :: IO ()
 main = do
     testToJsonConfigOutput
-    let config = (productionB <> prefixB "logs")
     bracket_
-        (buildAndSetupYamlLogging config testLoggerConfigPath)
+        (buildAndSetupYamlLogging productionB testLoggerConfigPath)
         removeAllHandlers
```
chshersh: I forgot about one thing: users of `log-warper` **MUST** call `removeAllHandlers` function at the end of logging action. Could you please move `bracket_` to `launch*` functions?

src/System/Wlog/Launcher.hs
```diff
@@ -12,44 +12,49 @@
...
 -- @node@ and @node.comm@.
 
 module System.Wlog.Launcher
        ( buildAndSetupYamlLogging
-       , initLoggingFromYaml
        , parseLoggerConfig
-       , runLoggingWithFile
-       , runDefaultLogging
+       , launchFromFile
+       , defaultConfig
```
chshersh: Exports are not sorted alphabetically.

src/System/Wlog/Launcher.hs
```diff
@@ -123,46 +128,30 @@ buildAndSetupYamlLogging configBuilder loggerConfigPath = do
...
+    usingLoggerName loggerName action
+
+-- | Default logging configuration with the given 'LoggerName'.
+-- The root logger is set up with severity 'System.Wlog.Severity.Warning' and upper, the severity
+-- for given logger is set to 'System.Wlog.Severity.Debug' and upper, also time is shown in the log messages,
+-- and console output is on.
+defaultConfig :: LoggerName -> LoggerConfig
+defaultConfig loggerName = fromScratch $ do
+    lcShowTime      .= Any True
+    lcShowTid       .= Any False
```
chshersh: You don't need to set `lcShowTid` because `Any False ≡ mempty`.

chshersh (changes requested):

PR #70, opened 2017-12-07
[#55] Remove _lcFilePrefix

### LINKED ISSUE #55: Remove 'filePrefix' from logger config
Because now you can specify directories in path for output of logger and directories will be created automatically, there's no actual need in `filePrefix` parameter. I think things can be simplified here:

https://github.com/serokell/log-warper/blob/eae120ac22bcb2397eea2b89929797d0b92f5a26/src/System/Wlog/LoggerConfig.hs#L240

vrom911:
Resolves #55

### DIFF
--- a/src/System/Wlog/Launcher.hs
+++ b/src/System/Wlog/Launcher.hs
@@ -42,12 +42,9 @@ import Control.Exception (throwIO)
 import Control.Lens (zoom, (.=), (?=))
 import Data.Time (UTCTime)
 import Data.Yaml (decodeFileEither)
-import System.Directory (createDirectoryIfMissing)
-import System.FilePath ((</>))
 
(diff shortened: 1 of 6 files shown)


src/System/Wlog/Launcher.hs
```diff
@@ -79,10 +78,9 @@ setupLogging mTimeFunction LoggerConfig{..} = do
                             _lcTermSeverityOut
                             _lcTermSeverityErr
 
-    liftIO $ setPrefix _lcFilePrefix
     processLoggers mempty _lcTree
   where
-    handlerPrefix = _lcFilePrefix ?: "."
+    handlerPrefix = "."
```
chshersh: I think you can remove `handlerPrefix` safely.

chshersh (changes requested):

chshersh (approved):

PR #72, opened 2017-12-13
Upgrade universum to 0.9.1

vrom911:


src/System/Wlog/LoggerConfig.hs
```diff
@@ -241,6 +244,22 @@ data LoggerConfig = LoggerConfig
...
+        , _lcTermSeverityErr = orCombiner  _lcTermSeverityErr
+        , _lcShowTime        = andCombiner _lcShowTime
+        , _lcShowTid         = andCombiner _lcShowTid
+        , _lcConsoleAction   = andCombiner _lcConsoleAction
+        , _lcMapper          = andCombiner _lcMapper
+        , _lcTree            = andCombiner _lcTree
+        }
+      where
+        orCombiner  field = field lc1 <|> field lc2
+        andCombiner field = field lc1  <> field lc2
```
chshersh, line 263: Could you please move these two functions at top level and use them for `LoggerTree` `Semigroup` instance implementation? I propose to have names `appendBy` (or `andBy`) and `altBy` (or `orBy`). You can still have functions inside `where` though as partial specialisations of those two.

chshersh (changes requested):

src/System/Wlog/LoggerConfig.hs
chshersh: Though... I just thought that you can just duplicate code. It will be shorter.

stack-8.2.2.yaml
```diff
@@ -1,6 +1,9 @@
 resolver: nightly-2017-11-24
```
chshersh: @vrom911 I'm not sure why build failed. But probably this is because `stack-8.2.2.yaml` uses old `nigthly` snapshot. Could you try to change it to `importify` version?

https://github.com/serokell/importify/blob/e29429faed84ea104f064ac2d4fbd27aa5993ea2/stack-8.2.2.yaml#L1

I'm not sure this will fix... Because it worked earlier with `stack-8.2.2.yaml`.

chshersh (approved):

PR #82, opened 2017-12-21
[LW-26] Write Pure logging how-to

vrom911:


examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+```bash
+$ stack exec pure-how-to
+[counter:DEBUG] [2017-12-21 13:13:18.30 UTC] Before incrementation 0
+[counter:INFO] [2017-12-21 13:13:18.30 UTC] After incrementation 2
+
+```
+
+
+
+[pure]: https://github.com/serokell/log-warper/blob/master/src/System/Wlog/PureLogging.hs
```
chshersh: Maybe it's better to give link to hackage here as well?

http://hackage.haskell.org/package/log-warper-1.8.2/docs/System-Wlog-PureLogging.html

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+import qualified Data.Text as Text
+
+```
+
+## Pure code
+
+So, basically, if you want to log something in your function, but it doesn’t use
+the `IO` monad, you need to get some monad that implements `WithLogger`
+in your monad stack, so add the constraint and then use [`launchNamedPureLogWith`][lnplw].
+But let's have a closer look at that.
```
chshersh: I think it's not completely clear what person might want and what he can do with `log-warper`. It's not enough to have just pure function to use pure logging of `log-warper`. Logging is monadic in `log-warper`. So you can only add it if you're working in some monad. I would it rephrase it into something like this:

> Consider pure monadic action where you want to add logging to (sorry for my English). You can use `putStrLn` or something like that because you don't have `IO` in your type. But you still want to be able to inspect intermediate states of your computation. Thus you need some ability to log some information.

Not sure about correct wording... But I wouldn't mention `launchNamedPureLogWith` and `WithLogger` implementation details here. It's better to have high overview of problem.

TODO: maybe add some real-life example where you might want to do this and why?

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+
+Let's have this situation to give an example of how and where it should be used.
+Imagine that we have a newtype `PureCounter` which is actually wrapper on `State Int`.
```
chshersh: > which is actually wrapper on `State Int`.

I would add _where `Int` represents some accumulator.

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+## Pure code
+
+So, basically, if you want to log something in your function, but it doesn’t use
+the `IO` monad, you need to get some monad that implements `WithLogger`
+in your monad stack, so add the constraint and then use [`launchNamedPureLogWith`][lnplw].
+But let's have a closer look at that.
+
+Let's have this situation to give an example of how and where it should be used.
```
chshersh: Maybe rephrase like _Let's have some simple pure monad for our simple example_?

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+Imagine that we have a newtype `PureCounter` which is actually wrapper on `State Int`.
+
+```haskell
+
+newtype PureCounter a = PureCounter
+    { runPureCounter :: State Int a
+    } deriving (Functor, Applicative, Monad)
+```
+
+Also we have a class 'MonadCounter' where it will all work in.
```
chshersh: Backticks should be used here instead of '' in 'MonadCounter'

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+With this class we will be able to get the current counter
+and to increase the counter by one
+
+```haskell
+
+class Monad m => MonadCounter m where
+    getCounter :: m Int
+    incCounter :: m Int
```
chshersh: Maybe add comment like `-- returns resulting counter after incrementation`?

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+
+```
+
+Let's implement the instance for our data type
+
+```haskell
+
+instance MonadCounter PureCounter where
+    getCounter = PureCounter $ get
+    incCounter = PureCounter $ modify (+ 1) >> get
```
chshersh: Let's use `modify'` here instead of `modify`. It's not important for tutorial. But in general `modify'` should be used in such cases and I think it's better to use proper function.

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+
+```
+
+Also some useful functions
```
chshersh: I would split this section into two. Like

> We need a way to evaluate our pure computation, so let's add such function.

and the second one is

> Let's have function which works only for `MonadCounter` for testing purposes

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+evalPureCounter :: PureCounter a -> a
+evalPureCounter pc = evalState (runPureCounter pc) 0
+
+doubleInc :: MonadCounter m => m Int
+doubleInc = incCounter >> incCounter
+
+```
+
+That `doubleInc` seems to be good candidate to add some logging, cause
+it's doing some smart stuff and it would be great to have that info logged properly.
```
chshersh, line 105: Nice! I like how you said about smart stuff :D I also think this function is too difficult and should be logged.

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+    counterAfter <- doubleInc
+    logInfo $ "After incrementation " <> Text.pack (show counterAfter)
+
+```
+You see how we added [`logDebug`][logDebug] and [`logInfo`][logInfo]
+functions into our code. But these functions demands `m` to be in
+class `WithLogger`.
+
+Now we want to run our pure function. Let's add this function
+that will be called from `main`.
```
chshersh: Maybe add that in order to actually observe result result of logging we need to print the result inside `IO` monad thus we need to have constraint `WithLoggerIO` here?

examples/PureLogging.md
```diff
@@ -0,0 +1,157 @@
...
+
+```
+
+Here you see [`launchNamedPureLogWith`][lnplw] which is basically
+makes logging happen. It gathers all the logs that were written in
+`doubleIncWithLog` and will take them out of there.
+
+Also take into consideration that along  with `launchNamedPureLogWith` there is [`launchNamedPureLog`][lnpl] for your needs.
+
+But for that we will need instance of `MonadCounter` for `NamedPureLogger`
```
chshersh: I think it's better to say like this:

> To be able to launch pure logging we also need to implement instance of `MonadCounter` for `NamedPureLogger` data type from `log-warper` library. This data type is responsible for collecting logs. Hopefully, implementing instances for this data type is rather easy!

chshersh (changes requested):
Looks really good! I really like your example and how you described everything step by step. But I think that more clarifications is required. And we need to provide code with more details. See my comments.

examples/PureLogging.md
vrom911: I didn't found good real-life example to describe in here.. :disappointed:

chshersh (approved):

PR #83, opened 2017-12-22
[LW-4] Add Concurrent module

### LINKED ISSUE #52: [LW-4] Add logging in parallel with long action function
Details are here: https://issues.serokell.io/issue/LW-4

vrom911:
Resolves #52

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+1.8.7
+=====
+
+* [#52](https://github.com/serokell/log-warper/issues/52):
+  Add `System.Wlog.Concurrent` module.
+
 1.8.6
 =====
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,8 @@
 =====
 
 * [#52](https://github.com/serokell/log-warper/issues/52):
-  Add `System.Wlog.Concurrent` module.
+  Add `System.Wlog.Concurrent` module for `ghc-8.2.2` which
+  allows to run action in parallel with logging.
 
 1.8.6
 =====
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -1,15 +1,29 @@
+{-# LANGUAGE CPP                 #-}
+{-# LANGUAGE ExplicitForAll      #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
 -- | Testing module to play with logging.
 
 module Main where
 
 import Universum
 
+#if ( __GLASGOW_HASKELL__ >= 802 )
+import Control.Monad.Trans.Control (MonadBaseControl)
+#endif
 import Data.Yaml.Pretty (defConfig, encodePretty)
 import Lens.Micro ((?~))
+#if ( __GLASGOW_HASKELL__ >= 802 )
+import Time (sec, threadDelay)
+#endif
 
 import System.Wlog (CanLog, atLogger, defaultConfig, infoPlus, launchFromFile, launchWithConfig,
                     logDebug, logError, logInfo, logNotice, logWarning, ltSeverity,
                     modifyLoggerName, parseLoggerConfig, productionB, usingLoggerName)
+#if ( __GLASGOW_HASKELL__ >= 802 )
+import System.Wlog (WithLoggerIO, launchSimpleLogging, logWarningWaitInf)
+#endif
 
 testLoggerConfigPath :: FilePath
 testLoggerConfigPath = "logger-config-example.yaml"
@@ -52,3 +66,15 @@ main = do
 
     putTextLn "\nFrom file configurations.."
     launchFromFile testLoggerConfigPath "node" runPlayLog
+
+#if ( __GLASGOW_HASKELL__ >= 802 )
+    launchSimpleLogging "concurrent" concurrentActions
+
+concurrentActions :: forall m . (WithLoggerIO m, MonadBaseControl IO m) => m ()
+concurrentActions = logWarningWaitInf 2 "stupid action" someStupidAction
+  where
+    someStupidAction :: m ()
+    someStupidAction = replicateM_ 10 $ do
+      threadDelay $ sec 1
+      liftIO $ putTextLn "HEYYEYAAEYAAAEYAEYAA"
+#endif
--- a/src/System/Wlog/Concurrent.hs
+++ b/src/System/Wlog/Concurrent.hs
@@ -1,0 +1,94 @@
+{-# LANGUAGE ConstraintKinds     #-}
+{-# LANGUAGE ExplicitForAll      #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module System.Wlog.Concurrent
+       ( WaitingDelta (..)
+       , CanLogInParallel
+       , logWarningLongAction
+       , logWarningWaitOnce
+       , logWarningWaitLinear
+       , logWarningWaitInf
+       ) where
+
+import Universum
+
+import Control.Concurrent (threadDelay)
+import Control.Concurrent.Async.Lifted (withAsyncWithUnmask)
+import Control.Monad.Trans.Control (MonadBaseControl)
+import Data.Time.Units (Microsecond, Second, convertUnit, toMicroseconds)
+import Fmt
+
+import System.Wlog.CanLog (WithLoggerIO, logWarning)
+
+-- | Data type to represent waiting strategy for printing warnings
+-- if action take too much time.
+data WaitingDelta
+    = WaitOnce      Second              -- ^ wait s seconds and stop execution
+    | WaitLinear    Second              -- ^ wait s, s * 2, s * 3  , s * 4  , ...      seconds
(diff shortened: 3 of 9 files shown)


CHANGES.md
```diff
@@ -1,3 +1,9 @@
+1.8.4
+=====
+
+* [#52](https://github.com/serokell/log-warper/issues/52):
+  Add `System.Wlog.Concurrent` module.
```
chshersh: Maybe tell here what functionality this module contains? Like, it allows to run action in parallel with logging or something like that.

src/System/Wlog/Concurrent.hs
```diff
@@ -0,0 +1,94 @@
...
+       , logWarningWaitInf
+       ) where
+
+import Universum
+
+import Control.Concurrent (threadDelay)
+import Control.Concurrent.Async.Lifted (withAsyncWithUnmask)
+import Control.Monad.Trans.Control (MonadBaseControl)
+import Data.Time.Units (Microsecond, Second, convertUnit, toMicroseconds)
+import Fmt
```
chshersh: Could you please make this import explicit?

stack.yaml
```diff
@@ -1,6 +1,7 @@
 resolver: lts-10.0
 packages: [.]
-extra-deps: []
+extra-deps:
+- time-units-1.0.0
```
chshersh: Unfortunately, `time-units` package is not in LTS :( Adding this `extra-dep` won't allow us to be in LTS anymore which is sad... So let's remove this dependency. And use `Integer` data type instead of type safe `time-units`

src/System/Wlog/Concurrent.hs
gromakovsky: Do we permit implicit imports nowadays? 🤔

chshersh (changes requested):
One more humble request: could you please add example into `play-log` which shows this feature? Like, you have dummy action, which waits 2.5 seconds, logs some `Info` message and then waits 2.5 second more. And using `logWarningLongAction` function you need to run in parallel with this action logger which warns every second about action being too long.

stack.yaml
vrom911: Probably we could make newtypes instead, because it is easy to mess up with the same type but different meanings? :thinking:

chshersh: @vrom911 You're 146% right. It's really a shame that such package like `time-units` is not on stackage... Having newtypes for such problem is the right solution. I agree with you here, let's introduce our own newtypes.

vrom911: I would like to wait the results of the discussion in slack about what to do with issue before applying newtypes.. Newtypes is not the best option in this case.. :(

gromakovsky: Ping.

chshersh: @gromakovsky We started work on `safe-time` library. Join #safe-time channel in Slack.

src/System/Wlog/Concurrent.hs
```diff
@@ -56,39 +57,43 @@ logWarningLongAction logFunc delta actionTag action =
     -- this function is going to be called under 'mask'.
     withAsyncWithUnmask (\unmask -> unmask $ waitAndWarn delta) (const action)
   where
-    printWarning :: Second -> m ()
+    printWarning :: (Time Second) -> m ()
```
chshersh: `()` in type are redundant here and in other places

src/System/Wlog/Concurrent.hs
```diff
@@ -56,39 +57,43 @@ logWarningLongAction logFunc delta actionTag action =
...
+                let newAcc    = acc + delayT
+                let newDelayT = k `timeMul` delayT
+                delayAndPrint delayT (toUnit @Second newAcc)
+                waitLoop newAcc newDelayT
+        in waitLoop 0 ms
 
-    delayAndPrint s t = do
-        liftIO $ threadDelay $ fromInteger $ toMicroseconds s
-        printWarning t
+    delayAndPrint delayT printT = do
```
chshersh: Could you please add type signature to this function?

stack.yaml
```diff
@@ -1,6 +1,6 @@
 resolver: lts-10.3
 extra-deps:
-- time-units-1.0.0
+- o-clock-0.0.0
```
chshersh: `o-clock` actually is on Stackage. Could you please add one more `stack` file called `stack-nightly.yaml` for this resolver:

* https://www.stackage.org/nightly-2018-01-29

And you can remove `extra-deps` from this `.yaml` file because this resolver contains `universum-1.0.4` and `o-clock`. Also would be good if you add this `.yaml` file to Travis CI.

src/System/Wlog/Concurrent.hs
```diff
@@ -14,31 +14,32 @@ module System.Wlog.Concurrent
...
 data WaitingDelta
-    = WaitOnce      Second              -- ^ wait s seconds and stop execution
-    | WaitLinear    Second              -- ^ wait s, s * 2, s * 3  , s * 4  , ...      seconds
-    | WaitGeometric Microsecond Double  -- ^ wait m, m * q, m * q^2, m * q^3, ... microseconds
+      -- | wait s seconds and stop execution
+    = WaitOnce      (Time Second)
+      -- | wait s, s * 2, s * 3  , s * 4  , ...      seconds
+    | WaitLinear    (Time Second)
+      -- | wait m, m * q, m * q^2, m * q^3, ... microseconds
+    | WaitGeometric (Time Microsecond) RatioNat
```
chshersh: You can store `Time Second` here as well. `Microsecond` was used here because multiplying integer seconds and some `Double` won't work.

chshersh (changes requested):

chshersh:
@vrom911 Oh, I forgot. Could you create simple usage example in playground?

vrom911:
@ChShersh you mean to take some example from `cardano-sl`?

chshersh:
@vrom911 No, I mean something simple. Like, create action where you

1. Wait 1 second.
2. Print something.

Repeat this action 10-15 times.

And run this action in parallel with geometric logging starting from 2 seconds.

Does this look clear to you?

chshersh (approved):

chshersh:
@gromakovsky Please, review one more time 🙏

CHANGES.md
```diff
@@ -1,3 +1,10 @@
+1.8.7
```
gromakovsky, line 1: I am not sure, but my understanding was that last component of version is only for fixes, but this is a new feature. So I think it would make more sense to call this version 1.9.0. But I don't mind leaving 1.8.7 actually. Up to you.

src/System/Wlog/Concurrent.hs
```diff
@@ -0,0 +1,99 @@
+{-# LANGUAGE ConstraintKinds     #-}
+{-# LANGUAGE ExplicitForAll      #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
```
gromakovsky, line 5: Having a Haddock comment here would be good.

examples/Playground.hs
```diff
@@ -52,3 +66,15 @@ main = do
 
     putTextLn "\nFrom file configurations.."
     launchFromFile testLoggerConfigPath "node" runPlayLog
+
+#if ( __GLASGOW_HASKELL__ >= 802 )
+    launchSimpleLogging "concurrent" concurrentActions
+
+concurrentActions :: forall m . (WithLoggerIO m, MonadBaseControl IO m) => m ()
+concurrentActions = logWarningWaitInf 2 "Second" someStupidAction
```
gromakovsky: `"Second"` is the name of `someStupidAction`. So there will be something like
> Action `Second` took more than…

Wht `"Second"`? Why not `stupid action`, for example? It looks confusing.

gromakovsky (approved):
I have some minor comments, but otherwise looks good. @ChShersh can you incorporate these changes into https://github.com/input-output-hk/cardano-sl/pull/2327 after this PR is merged and version is uploaded to Hackage?

CHANGES.md
chshersh: Citing @neongreen:

```
a.b.c.d

a = ALL-NEW AWESOME COOL, i.e. number useful for marketing basically
b = actually meaningful “this version definitely breaks API” number
c = “this version adds features but doesn't break API” number
d = “I fixed haddocks”, “I fixed compilation with GHC 7.6.2”, etc
```

chshersh:
>  @ChShersh can you incorporate these changes into input-output-hk/cardano-sl#2327 after this PR is merged and version is uploaded to Hackage?

Sure, I can.

examples/Playground.hs
vrom911: Oh, sorry there were two actions when I started to play with those functions and they were `First` and `Second`, and I forgot to change names to something that would make more sense.. Thanks for finding this out!

CHANGES.md
vrom911: Oh, I see. @ChShersh , what are your thought?

chshersh: @vrom911 I think `1.8.7` is okay since it doesn't break API.

gromakovsky: If we are using this convention for `log-warper`, then alright, `1.8.7` makes sense.

gromakovsky:
Let's merge and upload to Hackage maybe? 😉

chshersh:
Uploaded: http://hackage.haskell.org/package/log-warper

PR #84, opened 2017-12-25
Simplify launchSimpleLogging

dniku:


src/System/Wlog/Launcher.hs
```diff
@@ -225,7 +225,4 @@ launchSimpleLogging :: (MonadIO m, MonadMask m)
                     -> LoggerNameBox m a
                     -> m a
 launchSimpleLogging loggerName action =
-    bracket_
-        (setupLogging Nothing $ defaultConfig loggerName)
-        removeAllHandlers
-        (usingLoggerName loggerName action)
+  launchWithConfig (defaultConfig loggerName) loggerName action
```
chshersh: You can use η-reduce here to make `launchSimpleLogging` even shorter. And also write it on single line.

dniku: Fixed. However, note that the line is now 88 chars long and personally I'd prefer to split it.

vrom911 (approved):

src/System/Wlog/Launcher.hs
chshersh: Our style guide permits lines up to 100 characters. Well, if you want you might even write something shorter like this:


```haskell
launchSimpleLogging = flip launchWithConfig <*> defaultConfig
```

But you definitely shouldn't write such code. Current implement is nice.

Issue #21, opened 2017-10-06
Write good readme and maybe even tutorial


chshersh:
Current README doesn't explains how to use `log-warper` which is bad. This README should be written. Too much effort was put into `log-warper`. So we must try to make this library more production-ready.

chshersh:
Now we have `HowTo.md` with example of how to configure logging system and replace your `putStrLn` with `logInfo`. But this is not the whole tutorial. Here I will outline remaining parts:

## Plan

- [x] How To use `log-warper` in your project: resolved in PR #58 
- [x] How to use pure logging LHS and MD example
- [x] README.md updated with all features of `log-warper`
- [ ] Reference/FAQ section

In this message I will describe in more details what should be covered in above sections. This list might not be complete though...

## Features

`log-warper` provides next features:

1. Colored logging output.
2. Hierarchical logger names.
3. Different severity levels of messages with the ability to configure `Set` of severities.
4. All* (well, almost) configuration can be described in `.yaml` file.
5. Pure logging.
6. Flexible and easy creation of `LoggerConfig` using monoidal builders and lenses.
7. In-Memory logging with the ability to acquire last `N` megabytes of logs from in-memory cache.
8. Monadic interface for logging.
9. Logger rotation.

## FAQ

1. How can I redirect all output to `stderr`?
2. How can I disable only `Error` messages for my logger?
3. How can I show `ThreadId` inside log message?
4. How to easily disable terminal output?
5. How can I enable messages with severity `Info` and higher?
6. How can I log inside functions like `forkIO`?
7. How can I easily log exceptions without throwing them?
8. How should I configure logging: through code or through `.yaml` file?

PR #85, opened 2017-12-25
Fix README

gromakovsky:
Pure logging is based on `StateT` nowadays, at least that's what I saw
in code.


chshersh (approved):
You're right. There's open issue #21 under which whole README will be updated.

PR #88, opened 2017-12-30
#86 AtLogger lens

### LINKED ISSUE #86: Add lens for changing logger names
Sometimes it happens so that you want to change default logger configuration by changing properties of some logger. For example, in order to change default severity of specified logger you need to write something like this:

```haskell
defaultConfig myName & lcTree . ltSubloggers . at myName . _Just . ltSeverity ?~ mySeverity
```

This is a lot of code... It will be much more convenient, if `log-warper` provides additional helper lens

```haskell
atLogger :: LoggerName -> Traversal' LoggerConfig (Maybe LoggerTree)  -- not sure about type, didn't type check it
atLogger myName = lcTree . ltSubloggers . at myName . _Just  -- also not sure about name
```

Previous example can be rewritten in much shorter way using this lens:

```haskell
defaultConfig myName & atLogger myName . ltSeverity ?~ mySeverity
```

vrom911:
Resolves #86

### DIFF
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -5,10 +5,11 @@ module Main where
 import Universum
 
 import Data.Yaml.Pretty (defConfig, encodePretty)
+import Lens.Micro ((?~))
 
-import System.Wlog (CanLog, defaultConfig, launchFromFile, launchSimpleLogging, logDebug, logError,
-                    logInfo, logNotice, logWarning, modifyLoggerName, parseLoggerConfig,
-                    productionB, usingLoggerName)
+import System.Wlog (CanLog, atLogger, defaultConfig, infoPlus, launchFromFile, launchWithConfig,
+                    logDebug, logError, logInfo, logNotice, logWarning, ltSeverity,
+                    modifyLoggerName, parseLoggerConfig, productionB, usingLoggerName)
 
 testLoggerConfigPath :: FilePath
 testLoggerConfigPath = "logger-config-example.yaml"
@@ -44,9 +45,10 @@ main = do
     testToJsonConfigOutput
     let runPlayLog = testLogging >> showSomeLog
 
-    putStrLn $ encodePretty defConfig $ defaultConfig "example"
-    putTextLn "Default configurations.."
-    launchSimpleLogging "node" runPlayLog
+    putTextLn "Default configurations with modification.."
+    launchWithConfig (defaultConfig "node" & atLogger "node" . ltSeverity ?~ infoPlus)
+                     "node"
+                     runPlayLog
 
     putTextLn "\nFrom file configurations.."
     launchFromFile testLoggerConfigPath "node" runPlayLog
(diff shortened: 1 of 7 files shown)


src/System/Wlog/LoggerConfig.hs
```diff
@@ -351,3 +354,24 @@ maybeLogsDirB prefix = mempty { _lcLogsDirectory = prefix }
...
+--     . 'ltSeverity' ?~ 'warningPlus' )
+--     "myLogger"
+-- @
+--
+atLogger :: LoggerName -> Traversal' LoggerConfig LoggerTree
+atLogger logName = lcTree . leveldown (LoggerName <$> Text.splitOn "." (getLoggerName logName))
+  where
+    leveldown []     = error "Logger should be provided"
+    leveldown [x]    = ltSubloggers . at x . _Just
+    leveldown (x:xs) = ltSubloggers . at x . _Just . leveldown xs
```
chshersh: Could you add type to this function? Maybe `-XPartialSignatures` can help you...

src/System/Wlog/LoggerConfig.hs
```diff
@@ -351,3 +354,24 @@ maybeLogsDirB prefix = mempty { _lcLogsDirectory = prefix }
...
+--     & 'atLogger' "myLogger"
+--     . 'ltSeverity' ?~ 'warningPlus' )
+--     "myLogger"
+-- @
+--
+atLogger :: LoggerName -> Traversal' LoggerConfig LoggerTree
+atLogger logName = lcTree . leveldown (LoggerName <$> Text.splitOn "." (getLoggerName logName))
+  where
+    leveldown []     = error "Logger should be provided"
+    leveldown [x]    = ltSubloggers . at x . _Just
```
chshersh: I think we can move `ltSubloggers . at x . _Just` out to avoid code duplication.

src/System/Wlog/LoggerConfig.hs
```diff
@@ -351,3 +354,24 @@ maybeLogsDirB prefix = mempty { _lcLogsDirectory = prefix }
...
+--     ( defaultConfig "myLogger"
+--     & 'atLogger' "myLogger"
+--     . 'ltSeverity' ?~ 'warningPlus' )
+--     "myLogger"
+-- @
+--
+atLogger :: LoggerName -> Traversal' LoggerConfig LoggerTree
+atLogger logName = lcTree . leveldown (LoggerName <$> Text.splitOn "." (getLoggerName logName))
+  where
+    leveldown []     = error "Logger should be provided"
```
chshersh: Could you please `logName` to error `message`? Btw, here you should use `bug` from `universum` instead of `error`.

Another possible solution is to return `Nothing` when there's no such logger name. 

Which one do you prefer more, @vrom911?

examples/Playground.hs
```diff
@@ -44,9 +45,10 @@ main = do
     testToJsonConfigOutput
     let runPlayLog = testLogging >> showSomeLog
 
-    putStrLn $ encodePretty defConfig $ defaultConfig "example"
-    putTextLn "Default configurations.."
-    launchSimpleLogging "node" runPlayLog
+    putTextLn "Default configurations with modification.."
+    launchWithConfig (defaultConfig "node" & atLogger "node" . ltSeverity ?~ infoPlus)
+                     "node"
+                     runPlayLog
```
chshersh, line 51: Nice example! :+1:

chshersh (changes requested):

chshersh (approved):

Issue #89, opened 2018-01-02
Build failure in version 1.8.0


erikd:
```
src/System/Wlog/LoggerName.hs:40:15: error:
    • No instance for (Formatting.Buildable.Buildable LoggerName)
        arising from a use of ‘build’
    • In the expression: build
      In an equation for ‘loggerNameF’: loggerNameF = build
```

chshersh:
@erikd Could you elaborate more on the way you build `log-warper`? I can build `log-warper-1.8.0` without problems. Which building tool are you using? What are dependencies versions?

erikd:
I'm using the [mafia](https://github.com/haskell-mafia/mafia) build tool. Mafia is a thin wrapper around cabal.

I can see from your `.tavis.yml` file that you only test with Stack and from your cabal file that you only specify lower bounds. This leaves dependencies somewhat under constrained and it is therefore possible to make this un-buildable with cabal (and hence mafia).

Let me do some more investigation.

erikd:
Its pretty trivial to reproduce this just using cabal:
```
git clone https://github.com/serokell/log-warper
cd log-warper
git checkout v1.8.0 -b branch-v1.8.0
cabal sandbox init
cabal install --dependencies-only --enable-tests
cabal build
```
which results in:
```
src/System/Wlog/LoggerName.hs:40:15: error:
    • No instance for (Formatting.Buildable.Buildable LoggerName)
        arising from a use of ‘build’
    • In the expression: build
      In an equation for ‘loggerNameF’: loggerNameF = build
```
For packages that are destined to be uploaded to Hackage, I would actually recommend against using Stack to build them in Travis, for exactly this reason.

I've you're interested, I could modify your `.travis.yml` file to use Cabal instead of Stack and then fix the version bounds.

erikd:
Building the `v1.8.0` tag required the following patch to the cabal file:
```
diff --git a/log-warper.cabal b/log-warper.cabal
index 08ccf56..0da2144 100644
--- a/log-warper.cabal
+++ b/log-warper.cabal
@@ -51,7 +51,7 @@ library
                      , extra                >= 1.4.10
                      , filepath             >= 1.4.1
                      , fmt                  >= 0.5.0.0
-                     , formatting           >= 6.2.2
+                     , formatting           >= 6.2.2        && < 6.3
                      , hashable             >= 1.2.4.0
                      , lens                 >= 4.14
                      , monad-control        >= 1.0.1.0
[...]

PR #91, opened 2018-01-04
[#89] Specify upper bounds in .cabal file

chshersh:


vrom911 (approved):
:+1:

mihaimaruseac:
Note that because this PR introduces `^>` syntax in the Cabal file and that one is unsupported on Stackage build system, `log-warper-1.8.4` will be the version that Stackage uses for a period of time (until either Stackage can support the new syntax or the cabal file is reverted to not use bleeding-edge syntax).

chshersh:
@mihaimaruseac Could you please elaborate more on this? I can use `stack` locally to build `log-warper`, I have not problems with this new syntax. I'm using `stack-1.6.3`.

.travis.yml
```diff
@@ -1,6 +1,5 @@
-language: generic
-
-sudo: false
+sudo: true
```
int-index, line 1: Why did you do this? Sudoless container-based infrastructure on Travis is faster.

.travis.yml
```diff
@@ -1,6 +1,5 @@
...
+language: haskell
```
int-index, line 2: And why did you do this? It makes Travis install a GHC compiler, which we don't want (we do it manually).

.travis.yml
```diff
@@ -16,21 +15,24 @@ matrix:
 
   include:
 
-  - env: STACK_YAML="$HOME/build/serokell/log-warper/stack.yaml"
+  - ghc: 8.2.2
```
int-index, line 18: This is redundant, the compiler version is determined by the stack file.

.travis.yml
```diff
@@ -16,21 +15,24 @@ matrix:
...
+    env: STACK_YAML="$HOME/build/serokell/log-warper/stack.yaml"
     os: linux
     addons:
       apt:
         packages:
         - libgmp-dev
 
-  - env: STACK_YAML="$HOME/build/serokell/log-warper/stack-8.0.2.yaml"
+  - ghc: 8.0.2
```
int-index, line 26: This is redundant, the compiler version is determined by the stack file.

.travis.yml
```diff
@@ -16,21 +15,24 @@ matrix:
...
+    env: STACK_YAML="$HOME/build/serokell/log-warper/stack-8.0.2.yaml"
     os: linux
     addons:
       apt:
         packages:
         - libgmp-dev
 
-  - env: GHCVER='8.2.2' CABALVER='head'
+  - ghc: 8.2.2
```
int-index, line 34: This is redundant, the compiler version is determined by GHCVER here.

int-index (commented):
Can you undo Travis file changes? I don't see the motivation behind them, seem to make things worse (slower).

mihaimaruseac:
@ChShersh see https://github.com/commercialhaskell/stack/issues/3464

.travis.yml
chshersh: According to research of our DevOps team, `sudo: true` makes builds faster. If this it not the case, you should contact our DevOps team, I'm just using instructions from the article. Read Travis CI in Haskell article by our team.

.travis.yml
chshersh: According to our article on Travis CI in Haskell, setting `language` option to `Haskell` also makes builds faster. It's not documented in Travis CI, but `language: Haskell` contains GHC 8 and instead of installing it, Travis CI uses preinstalled one.

.travis.yml
chshersh: This is not redundant. Without this option it's harder to me to determine for which version of compiler CI is failing. Compare:

**without this option**

screenshot from 2018-01-05 14-58-05

**with this option**

screenshot from 2018-01-05 14-57-57

.travis.yml
chshersh: This is not redundant. Without this option it's harder to me to determine for which version of compiler CI is failing. Compare:

**without this option**

screenshot from 2018-01-05 14-58-05

**with this option**

screenshot from 2018-01-05 14-57-57

.travis.yml
chshersh: This is not redundant. Without this option it's harder to me to determine for which version of compiler CI is failing. Compare:

**without this option**

screenshot from 2018-01-05 14-58-05

**with this option**

screenshot from 2018-01-05 14-57-57

mihaimaruseac:
Stackage now supports the `^>` syntax, so [we've pulled in the last version of `log-warper`](https://www.stackage.org/diff/nightly-2018-02-21/nightly-2018-02-22).

PR #102, opened 2018-02-21
do not catch and squelch async exception

avieth:
This caused the following bug: a child thread, spawned by withAsync, and
linked to the main thread, threw an exception. The main thread was
presently trying to log something, so this simple logger grabbed the
async exception and tossed it out.

Solution: catch only IOExceptions.

This should also raise the question: if there's an exception in logging,
should we log it? i.e. should we have this exception handler at all? TBD

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+1.8.10
+======
+
+* Remove exception handler from simple logger, which would catch all exceptions,
+  even asynchronous ones.
+
 1.8.9
 =====
 
--- a/log-warper.cabal
+++ b/log-warper.cabal
@@ -1,5 +1,5 @@
 name:                log-warper
-version:             1.8.9
+version:             1.8.10
 synopsis:            Flexible, configurable, monadic and pretty logging
 homepage:            https://github.com/serokell/log-warper
 license:             MIT
--- a/src/System/Wlog/LogHandler/Simple.hs
+++ b/src/System/Wlog/LogHandler/Simple.hs
@@ -27,7 +27,6 @@ module System.Wlog.LogHandler.Simple
 import Universum
 
 import Control.Concurrent (modifyMVar_, withMVar)
-import Control.Exception (IOException)
 import Data.Text.Lazy.Builder as B
 import Data.Typeable (Typeable)
 import System.Directory (createDirectoryIfMissing)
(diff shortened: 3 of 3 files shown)


avieth:
> This should also raise the question: if there's an exception in logging,
should we log it? i.e. should we have this exception handler at all? TBD

What if there's an exception when logging the exception? And so on... I think the exception handler here should be dropped.

gromakovsky:
> What if there's an exception when logging the exception? And so on... I think the exception handler here should be dropped.

I checked `hslogger` and I see that this handler was there as well: https://github.com/jgoerzen/hslogger/blob/master/src/System/Log/Handler/Simple.hs#L70
I don't mind removing this exception handler altogether. I think the caller should decide what to do with such exceptions. And usually it means that the program is in a very bad state and probably should be stopped. So not catching exceptions should be ok.
cc @ChShersh what do you think?

chshersh:
@avieth @gromakovsky I'm also okay with removing this handler. Personally I've never encountered this error message. How to log exception while you're already logging something — is a tricky problem.

gromakovsky (changes requested):
Okay, so:
1. Let's remove this exception handling altogether (since three people agree it's a good idea).
2. Please bump version to 1.8.10 (is it right, @ChShersh?) and update `CHANGES.md`.

avieth:
Removed the handler, bumped version, updated CHANGES.md

chshersh (approved):
Looks good to me :+1: Though, I want to notice, that these changes won't be available in `cardano-sl` immediately because it `1.8.9` version I bumped up `universum` lower bound to `>= 1.1.0` during my work on CSL-2125:

* https://github.com/input-output-hk/cardano-sl/pull/2327

And this versions introduces a lot of changes to code, unfortunately. Probably if you need your changes urgently in `cardano-sl`, it's better to somehow move them on top of older commit...

avieth:
> And this versions introduces a lot of changes to code, unfortunately. Probably if you need your changes urgently in cardano-sl, it's better to somehow move them on top of older commit...

Yeah I just noticed :( ok, I'll back-port them. 1.8.0 is a good choice?

chshersh:
@avieth Yes, looks good.

gromakovsky (approved):

vrom911 (approved):

chshersh:
This change is implemented in `avieth/async_catch_1.8.0` branch. Here is the PR which brings this feature into CSL:

* https://github.com/input-output-hk/cardano-sl/pull/2566

I will use `log-warper-1.8.10` version (which also contains this bugfix) in CSL in my PR with migration to newer LTS to have this bugfix as well:

* https://github.com/input-output-hk/cardano-sl/pull/2327

PR #106, opened 2018-04-05
LW-34 Fix fallbacks for missing nodes

avnik:


chshersh (commented):
I reviewed. Logic is not simple, requires time to understand. Maybe it can be written simpler. But not sure. Otherwise looks good.

volhovm:
@ChShersh please consider reviewing https://github.com/serokell/log-warper/pull/107 instead.

PR #107, opened 2018-04-06
LW-34 Messages in nonexisting loggers are handled improperly

volhovm:


avnik:
LGTM :+1:

src/System/Wlog/IOLogger.hs
```diff
@@ -117,21 +118,24 @@ logInternalState = unsafePerformIO $ do
...
+Given a name, return all components of it, starting from the root.
+
 Example return value:
 
->["", "MissingH", "System.Cmd.Utils", "System.Cmd.Utils.pOpen"]
+λ> componentsOfName (LoggerName "a.b.c")
+[LoggerName {getLoggerName = ""},
+ LoggerName {getLoggerName = "a"},
+ LoggerName {getLoggerName = "a.b"},
+ LoggerName {getLoggerName = "a.b.c"}]
```
chshersh, line 130: Really nice example! I like to see documentation improvements :+1:

src/System/Wlog/IOLogger.hs
```diff
@@ -156,63 +160,67 @@ logMCond logname sev msg cond = do
...
+    parentLoggers :: LoggerName -> m [Logger]
+    parentLoggers = fmap reverse . mapM getLogger . componentsOfName
+
+    -- Tries to log the message into handlers. sevFiltPassed variable
+    -- denotes the "has log message passed through the first severity
+    -- filter". We only apply severity filter once, the first time we
+    -- encounter it.
+    traverseAndLog :: Bool -> [Logger] -> m ()
+    traverseAndLog _ [] = pass
+    traverseAndLog sevFiltPassed (x:xs) = do
```
chshersh: [`whenNotNull`](https://github.com/serokell/universum/blob/eec7d0dddd8ead098f6b177c2222a3d2475b9119/src/Universum/List/Safe.hs#L64-L66) from `universum` can be used here

src/System/Wlog/IOLogger.hs
```diff
@@ -156,63 +160,67 @@ logMCond logname sev msg cond = do
...
+        let doLog n = do
+                forM_ (x ^. lHandlers) callHandler
+                traverseAndLog n xs
+        case x ^. lLevel of
+           _         | sevFiltPassed -> doLog sevFiltPassed
```
chshersh: I want to say that this `case-of` expression is written in obscure way.

src/System/Wlog/IOLogger.hs
```diff
@@ -156,63 +160,67 @@ logMCond logname sev msg cond = do
...
+           Nothing   -> doLog sevFiltPassed
+           (Just lp) | sev `Set.member` lp -> doLog True
```
chshersh: For example, here you could use `when` function instead of guards.

src/System/Wlog/IOLogger.hs
volhovm: ~fixed

src/System/Wlog/IOLogger.hs
volhovm: Added.

volhovm:
@ChShersh CI doesn't pass, but it also doesn't in master (the failure is not related to this change), so can we merge it?

PR #108, opened 2018-04-20
[LW-36] v2.0: Redesign log-warper

chshersh:
New version of `log-warper`. Comes under version `2.0`.

`log-warper` has been split into two packages:

* `log-warper-core`: extremely low number of dependencies, only basic type classes and data types.
* `log-warper`: implementation of those interfaces for hierarchical logging just as before.

Not all parts are finished. But I just want to share this work to agree on final design.

There's [`v1`](https://github.com/serokell/log-warper/tree/v1) branch which is `log-warper` snapshot. That branch should be used for maintaining older version of `log-warper`. And, hopefully, will be removed in future.


log-warper-core/src/Log/Core/Class.hs
```diff
@@ -0,0 +1,201 @@
...
+'withLogAction' f . 'withLogAction' g ≡ 'withLogAction' (f . g)
+@
+
+* __@PutGet:@__ 'askLogAction' after 'withLogAction' should return modified action.
+
+@
+'withLogAction' f 'askLogAction' ≡ 'fmap' f 'askLogAction'
+@
+-}
+class Monad m => MonadLogger (m :: Type -> Type) (r :: Type) | m -> r where
```
int-index: Why was this order of type parameters chosen? I usually expect monad classes to have `m` as the last parameter.

log-warper-core/src/Log/Core/Class.hs
```diff
@@ -0,0 +1,201 @@
...
+import Data.Kind (Type)
+
+import Log.Core.Action (LogAction (..), cmap)
+
+{- | Type class for monads that have logging ability. In order to log things
+monad @m@ should be able to construct 'LogAction' for some logging record @r@.
+
+__Laws:__
+
+Some laws duplicate @lens@ laws. And maybe of the laws are redundant and can be
```
int-index: "maybe of the laws"? Did you skip a word?

log-warper-core/src/Log/Core/Class.hs
```diff
@@ -0,0 +1,201 @@
...
+'withLogAction' f 'askLogAction' ≡ 'fmap' f 'askLogAction'
+@
+-}
+class Monad m => MonadLogger (m :: Type -> Type) (r :: Type) | m -> r where
+    -- | Return 'LogAction' from current monadic context.
+    askLogAction :: m (LogAction m r)
+
+    default askLogAction :: (MonadTrans t, t n ~ m, MonadLogger n r)
+                         => m (LogAction m r)
+    askLogAction = do
```
int-index: I'm not a fan of `default` method implementations. The problem is that sometimes one might forget to implement some method and there will be no warning.

Instead I would export `askLogActionLifted` as a separate function from the library, and the user could simply write `askLogAction = askLogActionLifted`, explicitly specifying the behavior she wants.

log-warper-core/src/Log/Core/Class.hs
```diff
@@ -0,0 +1,201 @@
...
+import Control.Monad.Trans.Identity (IdentityT (..), mapIdentityT)
+import Control.Monad.Trans.Maybe (MaybeT (..), mapMaybeT)
+import Control.Monad.Trans.Reader (ReaderT (..), ask, mapReaderT)
+import Control.Monad.Trans.State.Strict (StateT (..), get, mapStateT)
+import Data.Kind (Type)
+
+import Log.Core.Action (LogAction (..), cmap)
+
+{- | Type class for monads that have logging ability. In order to log things
+monad @m@ should be able to construct 'LogAction' for some logging record @r@.
```
int-index: What does "logging record" mean? This documentation is not clear. From the definition of `LogAction`, I now think it's the message type.

What about renaming this parameter to `message`? It's what `logging-effect` does: https://hackage.haskell.org/package/logging-effect-1.2.5/docs/Control-Monad-Log.html#t:MonadLog

In fact, what is the advantage of this `MonadLogger` over the one defined in `logging-effect`?

log-warper-core/src/Log/Core/Class.hs
```diff
@@ -0,0 +1,201 @@
...
+    'logR' "Another logging"
+@
+-}
+withR :: MonadLogger m r => (r -> r) -> m a -> m a
+withR = withLogAction . cmap
+
+{- | Simple implementation for 'MonadLogger' type class. This data type just
+stores 'LogAction' which performs logging action under monad @m@.
+-}
+newtype ActionT r m a = ActionT
```
int-index: Now you're just copying `logging-effect` but with worse naming, aren't you?

chshersh: @int-index I never heard about `logging-effect` library before this comment, my bad. I've implemented these data types and type classes using my poor experience and your suggestions about monoidal actions. And I always appreciate better suggestions for naming!

log-warper-core/src/Log/Core/Class.hs
chshersh: @int-index Yes, I did. This should be "maybe some of the laws".

log-warper-core/src/Log/Core/Class.hs
chshersh: @int-index No particular reason. I also thought about this. Order can be changed easily.

log-warper-core/src/Log/Core/Class.hs
int-index: Maybe just take `logging-effect` as the basis? The definitions are very similar, and we shouldn't fracture the ecosystem.

log-warper-core/src/Log/Core/Class.hs
chshersh: @int-index Yeah, another problem with `default` implementations is that it doesn't work well with `{-#  MINIMAL #-}` pragma :disappointed: I like `default` implementations when instances can be one-liners. But in this case they can't, so probably having default function is better solution.

log-warper-core/src/Log/Core/Class.hs
chshersh: > I now think it's the message type

Yes, my English is poor.

> What about renaming this parameter to message?

This sounds like a good improvement!

> In fact, what is the advantage of this MonadLogger over the one defined in logging-effect?

As I mentioned earlier, I never heard about `logging-effect` library. Need to think about this.

log-warper/src/Log/Monad.hs
```diff
@@ -0,0 +1,120 @@
...
+* __@Idempotency:@__ context is not changed over time
+
+@
+askLogContext >> askLogContext ≡ askLogContext
+@
+
+Strictly speaking, this law doesn't hold for 'LoggerT' when somebody calls
+'Log.Core.Class.withLogAction' function but this should be ok.
+-}
+class Monad m => MonadLogContext m exts | m -> exts where
```
int-index: Assuming I write my code with this constraint, can I run it using `PureLoggerT`? If yes, I don't see how, there's no corresponding instance. If not, where's the class for this abstraction? I don't really understand how `PureLoggerT` and `LoggerT` are interchangeable.

log-warper/src/Log/Monad.hs
```diff
@@ -0,0 +1,120 @@
...
+type WarperAction exts m = LogAction (LoggerT exts m) (Event exts)
+
+{- | Type class for monads which have access to 'Context'.
+
+__Laws:__
+
+* __@Idempotency:@__ context is not changed over time
+
+@
+askLogContext >> askLogContext ≡ askLogContext
```
int-index: This law does not assure that the context is not changed over time. It only assures that `askLogContext` doesn't change itself.

I can say that for `StateT` we have:

```
get >> get = get
```

And yet there's `put`.

log-warper/src/Log/Monad.hs
```diff
@@ -0,0 +1,120 @@
...
+@
+
+Strictly speaking, this law doesn't hold for 'LoggerT' when somebody calls
+'Log.Core.Class.withLogAction' function but this should be ok.
```
int-index: I don't see how `withLogAction` can be used to change the logging context "over time", it can be used to override it locally.

log-warper/src/Log/Configuration/Extension.hs
```diff
@@ -0,0 +1,42 @@
...
+
+import Universum
+
+import Data.Type.Bool (If)
+
+-- | Kind for extension points which are used in logger configuration.
+data Extension = Rotation | Terminal
+
+-- | Type family for checking whether given 'Extension' is in list of extensions.
+type family ElemExtension (ext :: Extension) (exts :: [Extension]) :: Bool where
```
int-index: I don't 100% understand what those extensions are, but what is the motivation to track them on the type-level?

log-warper/src/Log/Monad.hs
chshersh: I hate `MonadLogContext` because it looks like hack and it's not the part of design I want to be. But I can't write it better :slightly_frowning_face: I didn't pay much attention to make `PureLoggerT` and `MonadLogContext` friends. But there should be no problem. The ideas behibd `launchPureLogging` functions which were previously in `log-warper` remain the same. Just need to uncomment those functions and rewrite them using new interface and you can use code with `MonadLogContext` with pure logging as well.

log-warper/src/Log/Monad.hs
chshersh: Yeah, I need to rephrase it and name this law as **`GetGet`**.

log-warper/src/Log/Monad.hs
chshersh: This sentence is nonsense. I just forgot to remove it.

log-warper/src/Log/Configuration/Extension.hs
chshersh: @int-index These extensions solve the following problem:

In `log-warper` we want to have single `.yaml` file which configures every part of logging. And with monoidal `LogAction` you need to perform two steps:

1. Specify options in `.yaml` file (for example, `rotation` options).
2. Don't forget to add `<> logRotation` to action you want to run.

So `Configuration` should be flexible in terms that it needs to support both presence of rotation paramenters and absence. Because I don't want this behavior: when you specify `rotation` parameters in `.yaml` file and don't see logger rotation. The idea is that when you add `<> logRotation` to resulting action, the type of `exts` type variable for `Configuration exts` should be infered as

```haskell
HasRotation exts => Configuration exts
```

And if you don't add `<> logRotation` but you specify `rotation` parameters in `.yaml` file you will see runtime parsing error. So, basically, this `Extension` mechanism allows you to extend `Configuration` data type by required parts of parameters and prevents your from accidentally forget to do one of both things: (1) specify parameters in `.yaml` or (2) add `<> logRotation`.

I'm not claiming that this solution is the best one. For example, we can remove this type-level magic and just pass `RotationParameters` explicitly to `logRotation`. But with this solution you now have some parts of configuration through `yaml` and some parts through code or multiple `yaml` files and I don't like any of alternative solutions.

Another solution: just as before. Store only `Maybe SomeParameters` in `Configuration`. And configure logging through boolean flags. But I don't really like this solution because it's too fragile and can be broken at any time (as you noticed earlier). I really like to replace:

```haskell
let myAction = logFile
```

with

```haskell
let myAction = logFile <> logStdout
```

to get stdout output instead of changing value of `.yaml` configuration.

int-index: I still don't understand why we have to track extensions at compile-time if the error will be triggered at runtime.

int-index: I'm trying to understand your example but there isn't `logRotation` defined anywhere in this PR (or GitHub hides it).

But I think I understand the basic problem you're dealing with. You want to compose loggers `l1 <> l2 <> l3` etc, but each can have its own section in the configuration YAML file, correct? Well, in this case you definitely can't afford to have a closed kind `data Extension = ...`, what if the user wants to define his own extension?

chshersh: @int-index I want the following behavior:

1. Parse **_successfully_** rotation parameters if they **are specified** and I **need** them.
2. **_Fail_** parsing when rotation parameters **are specified** and when I **don't need** them.
3. **_Fail_** on parsing when rotation parameters **aren't specified** but I **need** them.
4. Parse **_successfully_** when rotation parameters **aren't specified** and I **don't need** them.

There's supposed to be the only way with current design to tell that you **need** rotation parameters: `<> logRotation`. I don't know better way to require rotation parameters from configuration when you need to manually add additional action.

With previous design rotation was enabled only by `.yaml`. If you have rotation parameters you will have logging rotation. If you don't have rotation parameters you won't have logger rotation. The same design can be used with new interface as well if you like to remove all this type-level stuff. But instead of having on user side:

```haskell
let myAction = logFile <> logRotation
```

You will always have on user side the following:

```haskell
let myAction = logConfigurationAction
```

and `<> logRotation` will be performed internally in `log-warper` library, depending on whether you specified rotation parameters in `.yaml` file or not.

Just a matter of trade-off which design people like more and which behavior they want more and is it worth to have type-level magic.

chshersh: > Well, in this case you definitely can't afford to have a closed kind data Extension = ..., what if the user wants to define his own extension?

Yes, unfortunately :disappointed:  There're no open data kinds in GHC at the moment:
* https://ghc.haskell.org/trac/ghc/ticket/11080

But I didn't intend to provide such flexibility for user. It would be a good thing to do though (and I don't know how to do it). So for now this is only for library. In case we want to extend `log-warper` implementation with new ways of logging like `ElasticSearch` or something different we just patch `log-warper`. So, basically, we are the users of `log-warper` and if some feature required, new issue can be opened and new feature can be implemented.

I saw that Kiril did some work in `lootbox` regarding externally extensible configuration:
* https://github.com/serokell/lootbox/blob/master/code/config/lib/Loot/Config/Record.hs

I think if we want such thing it's better be moved into separate package so we can just replace my ad-hoc `Configuration` type with more general and more extensible solution (if it's possible).

The same thing applies to `Event` data type. Some cool tricks can be used to make extensible logging message type. But I think that such parts should be implemented independently and then logging library can just use those parts. For now I tried to replicate old behavior of `log-warper` with new design. In my defense I want to say with old `log-warper` it also wasn't possible to extend logging messages and logging actions.

int-index: >   Parse successfully rotation parameters if they are specified and I need them.
    Fail parsing when rotation parameters are specified and when I don't need them.
    Fail on parsing when rotation parameters aren't specified but I need them.
    Parse successfully when rotation parameters aren't specified and I don't need them.

Yes, I get this. What I don't get is why you need compile-time machinery for this check. I don't know what type `logRotation` in your example has, but it can be something like this:

```
(message -> m (), Set Extension)
```

then you'll have

```
logFile = (\_ -> writeFile ..., Set.empty)
logRotation = (\_ -> ..., Set.singleton Rotation)
```

When you combine these monoidally, the result will contain a `Set` with all _required_ configuration sections.

At the same time, after you parse a YAML file, you can easily get a `Set` with all _provided_ configuration sections.

Then you just compare these sets and throw an error if they don't match.

I'm not telling you to use this solution with `Set Extension` specifically, I'm just saying that I see no reason to make types more complicated for a check that is performed at runtime.

int-index: > In my defense 

It's not _your_ defense, we're just discussing code, relax :slightly_smiling_face:

int-index: >  Fail parsing when rotation parameters are specified and when I don't need them.

With the current approach, I don't really understand how you're going to achieve this. Okay, so doing `<> logRotation` will add a `HasRotation exts` constraint.

But in case I *don't* do `<> logRotation`, what will create a `HasNoRotation exts` constraint? Where will the `exts` list be specified in the end?

chshersh: > I'm not telling you to use this solution with Set Extension specifically, I'm just saying that I see no reason to make types more complicated for a check that is performed at runtime.

I agree this a good thing to consider. I didn't look at this problem form this angle. But I think there actually exist reason for such thing to be at compile time. And I will explain this below.

Logging message in new `log-warper` has the following type:

```haskell
data Event (exts :: [Extension]) = Event
    { _eName          :: !LoggerName
    , _eSeverity      :: !Severity
    , _eMessage       :: !Text
    , _eConfiguration :: !(Configuration exts)
    }
```

It stores logged data and whole configuration as well. I think that storing whole configuration in every logging message is convenient because every action don't need to have access to some monad with global context. It can just analyze fields of logging message. And with the help of  different combinators from `log-warper-core` every person can log things in some monadic block with changed `Configuration`.

So having:

```haskell
newtype LogAction m msg = LogAction { unLogAction :: msg -> m () }
```

I want the type of `logRotation` be the following (it's not implemented yet so probably it won't work at all):

```haskell
logRotation :: (HasRotation exts, MonadIO m) => LogAction m (Event exts)
```

It's intended that every logging action in `log-warper` should have form `LogAction m (Event exts)`.

So by adding `logFile <> logRotation` type of `exts` should satify `HasRotation` constraint.

The problem with checking such thing at runtime is that my global configuration probably will store something like this:

```haskell
data Configuration = Configuration
    { _cRotation :: Maybe RotationParameters
    ....
```

But if I add `logRotation` action I want to guarantee that `rotation` is always `Just` so I don't need to write `bug ImpossibleHappened` every time I pattern-match on `_cRotation`.

Currently this is achieved by the following type families:

```haskell
type HasRotation exts = ElemExtension 'Rotation exts ~ 'True
type ExtensionType ext exts param noExt = If (ElemExtension ext exts) param noExt
type RotationExtension exts = ExtensionType 'Rotation exts RotationParameters NoRotation

data NoRotation = NoRotation
```

And then in configuration I have:

```haskell
data Configuration (exts :: [Extension]) = Configuration
    { _cRotation :: RotationExtension exts
    ...
```

I'm not sure that having `HasRotation exts` constraint is enough to guarantee that type of `_cRotation` is actually inferred as `RotationParameters`. Maybe I need some trick you explained to Rinat with something like that:

```haskell
data MaybeN b a where
    JustN    :: a -> MaybeN True a
    NothingN :: MaybeN b a
```

> With the current approach, I don't really understand how you're going to achieve this.

I've already achieved this :slightly_smiling_face:  I tested this in playground and it works as expected. The solution is relatively simple. `_cRotation` field in `Configuration` will have either type `NoRotation` or `RotationParameters`. I just need to implement `FromJSON` instances for both types properly.

https://github.com/serokell/log-warper/blob/66417b8d03b526c4d7219afca074e0827884175c/log-warper/src/Log/Configuration/Rotation.hs#L71-L76

log-warper-core/src/Log/Core/Class.hs
```diff
@@ -0,0 +1,201 @@
...
+
+{- | Type class for monads that have logging ability. In order to log things
+monad @m@ should be able to construct 'LogAction' for some logging record @r@.
+
+__Laws:__
+
+Some laws duplicate @lens@ laws. And maybe of the laws are redundant and can be
+proven from other laws.
+
+* __@Idempotency:@__ @askLogAction@ shouldn't mutate monadic context:
```
volhovm: I think it's "Idempotence".

log-warper-core/src/Log/Core/Class.hs
volhovm: I would like to see them in the `r m` order too.

volhovm: So `MonadLogger m r` is basically equivalent to `MonadReader (LogAction m r) m`, right? Why did you decide to go with the explicit class declaration? 
(i may be missing something)

log-warper-core/src/Log/Core/Class.hs
volhovm: I am not sure that LW-core should depend on `logging-effect`. I took a short look at the library -- it does indeed resemble what is written here, but there are much more things that we may not want. It's up to Dima to decide. BTW `LoggingT` looks better to me.

log-warper/examples/Playground.hs
```diff
@@ -0,0 +1,110 @@
...
+main :: IO ()
+main = do
+    testToJsonConfigOutput
+
+    config <- fromRight (error "Parse error")
+          <$> decodeFileEither @(Configuration '[ 'Rotation ]) testLoggerConfigPath
+
+    launchLogger logStdout (Context "app" config) showSomeLog
+
+--     let runPlayLog = testLogging >> showSomeLog
```
volhovm: Unless we have proper tests, I want the lw-playground to be functional.

log-warper/src/Log/Configuration/Extension.hs
volhovm: 1. I don't like the previous LW configuration approach. It was overcomplicated -- mostly, because of the configuration-is-monoid approach, which I consider controversial.
2. User-defined extensions is a very neat feature and I think it should get into the release. Can we make a datatype per extension and a special typeclass which determines the extra action to be `mappend`ed?
3. I don't see any point in the attempt to lift configuration to the type level. I'd be satisfied with a dictionary (more or less as we had it in the original design), but with extension feature.

log-warper/src/Log/Configuration/Type.hs
```diff
@@ -0,0 +1,297 @@
...
+{- | Configuration for flexible and hierarchical logging. Logging configuration contains multiple parts:
+
+* 'GlobalParameters': description of logging parameters applied to every logger.
+* 'LoggerTree': per-logger configration.
+* 'Log.Configuration.Rotation.RotationParameters': parameters for logger rotation. Allowed to not be specified.
+
+Behavior of YAML parsing for any extension point (for example, 'Rotation') is
+the following:
+
+1. Has extension, has config fields: parses without problems.
```
volhovm: These four points can be summarized in "parsing succeeds iff extension and config fields are both present or both absent"

volhovm (commented):
This is only the first glance, I'll take a more detailed look at it tomorrow.

log-warper/src/Log/Configuration/Extension.hs
chshersh: > configuration-is-monoid approach

In new `log-warper` configuration is no longer a monoid.

> Can we make a datatype per extension and a special typeclass which determines the extra action to be mappended?

This is basically an _extensible records_ problem. It has various solutions. Maybe something like that can be implemented but I didn't put much thought into this.

log-warper-core/src/Log/Core/Class.hs
chshersh: I didn't think about this much. But `MonadLogger m r` is indeed equivalent to `MonadReader (LogAction m r) m`. Unfortunately, `MonadReader` type class has functional dependency. This means that in some big constraint alias you can't have:

```haskell
type WorkMode =
    ( MonanReader (LogAction m r) m
    , MonadReader SomethingElse m
    )
```

This won't compile. Which makes `MonadReader` approach less useful.

But what we can actually have is type class like `HasLogAction`. Though for now I'm not sure how it will look like...

log-warper-core/src/Log/Core/Class.hs
chshersh: @int-index @volhovm Okay, so I looked at `logging-effect` library. Approaches looks similar. But I still like my approach more :stuck_out_tongue_winking_eye:

`LoggingT` looks like better name to me as well. But I've already used `LoggerT` name inside `log-warper`. I found having both `LoggingT` and `LoggerT` confusing. So both of them should be renamed.

Here the reasons why I like my approach more and why I don't want to depend on `logging-effect`:

1. `logging-effect` has a huge number of dependencies we don't really need. If maintainer could move type class to something like `logging-effect-core` we could think on using this package as a basis for `log-warper`.
2. `MonadLog` type class from `logging-effect` looks much more complicated to me than `MonadLogger` from `log-warper-core`. I don't really understand this definition:

```haskell
class Monad m => MonadLog message m | m -> message where
    logMessageFree :: (forall n. Monoid n => (message -> n) -> n) -> m ()
```

  Well, what I understand perfectly is that I'm not a very smart developer. So maybe this is just my problem. But I'm also not the most stupid one. And from my experience I can tell that things that are easier to understand and grasp are also easier to reason about and it's easier to find bugs and implement instances and so on.

3. `Handler` in `logging-effect` is a type alias. `LogAction` is `log-warper-core` is a `newtype`. I like `newtypes` more. Type alias has its own benefits. But I like the following benefits of `newtype` more:

    * Ability to implement custom (and, as a consequence, more efficient) `Semigroup` and `Monoid` instances.
    * Type-safety? Anyone?..

4. Types of combinators for modifying logging action are fixed to specific `LoggingT` data type while in `log-warper-core`  they work with monad itself. Compare:

    * `logging-effect`:
        ```haskell
        mapLogMessageM :: MonadLog message' m 
                       => (message -> m message') 
                       -> LoggingT message m a 
                       -> m a
        ```
    * `log-warper-core`:
        ```haskell
        ward :: Applicative m => (a -> Bool) -> LogAction m a -> LogAction m a
        cbind :: Monad m => (a -> m b) -> LogAction m b -> LogAction m a
        withLogAction :: (LogAction m r -> LogAction m r) -> m a -> m a
        ```

  If I add more nested `mapLogMessageM` calls in `logging-effect` performance will be decreased because the more transformers you have on stack ( (and the bigger height of transoferms tower) the lower performance will be. With `log-warper-core` I don't add extra new transformers, I only apply functions to actions and I only change action which is stored inside monadic context.

5. I wrote big documentation and I don't want to throw it out...

log-warper/src/Log/Configuration/Extension.hs
volhovm: >This is basically an extensible records problem. 

Not really. This may sound too ambitious, but I think that there is no need to unite all "extensions" into one record. I can imagine the existential wrapper `SomeConf = forall x. HasNeededActionFunctionality x => SomeConf x` over some extension config (`RotationConf` or `StdoutConf`). We can read these from the file directly and then pass it to logger initializer. It may sound (and be) a complete bullshit, but does it at least seem to be a direction to think about? IIRC something similar was done in the old log-warper with handlers -- they were abstract, user-defined, abstracted into existentially qualified `HandlerT`.

log-warper/src/Log/Event.hs
```diff
@@ -0,0 +1,76 @@
...
+
+@
+'LogAction' someM ('Event' exts)
+@
+-}
+data Event (exts :: [Extension]) = Event
+    { _eName          :: !LoggerName
+    , _eSeverity      :: !Severity
+    , _eMessage       :: !Text
+    , _eConfiguration :: !(Configuration exts)
```
volhovm: I do not quite understand why is this field needed.

volhovm (commented):
The general low-level design is nice, though I still think we need to think more about extensions. The current implementation is not extensible and it also forces you to explicitly specify which extensions are expected in the config file. This, I think, is not what the regular user usually wants.

log-warper/src/Log/Event.hs
chshersh: @volhovm I explained why I think it's needed at the first half of this comment:

* https://github.com/serokell/log-warper/pull/108#discussion_r183084927

chshersh:
@int-index @volhovm I fixed code according to review comments.

Regarding _extensible configuration_: nice thing to have. But I propose to think about it in parallel with other changes. Personally I don't see fast an easy solution for this problem (since I'm not so experienced in type-level magic; existential constructors like for old `HandlerT` is not really good solution to me...). So I can open issues for other things that need to be done and then they can be done in parallel, maybe @vrom911 could take some issue or maybe somebody else want to hack on new `log-warper`.

chshersh:
Oh, and of course after I fix CI.

log-warper/src/Log/Event.hs
```diff
@@ -0,0 +1,76 @@
...
+in @log-warper@ should have form:
+
+@
+'LogAction' someM ('Event' exts)
+@
+-}
+data Event (exts :: [Extension]) = Event
+    { _eName          :: !LoggerName
+    , _eSeverity      :: !Severity
+    , _eMessage       :: !Text
```
avnik: Do we want `Text` here? I suggest to think about `Data.Text.Lazy.Builder` here, like `katip` use (https://github.com/Soostone/katip/blob/master/katip/src/Katip/Core.hs#L184). possible it allow make logging more flexible and not depending on Text type here (be polymorphic against Text/String/Bytestring/...)

log-warper/src/Log/Name.hs
```diff
@@ -0,0 +1,59 @@
...
+representation those commponents separated by dot character. Below you can see
+example of correspondence between list of components and string:
+
+@
+["foor", "bar", "baz"] <-> "foo.bar.baz"
+@
+-}
+newtype LoggerName = LoggerName
+    { unLoggerName :: Vector Text
+    } deriving (Eq, Ord, Generic, Semigroup, Monoid, Container)
```
avnik: `Vector`, why not list? (Just from performance perspective -- I noted IsList below)

avnik (commented):
General question -- can we finally drop hslogger derived GPL3 licensed stuff? (or at least concentrate it in one sub-package? I want to avoid gpl3 when possible)

README.md
```diff
@@ -7,98 +7,11 @@ log-warper
 [![Stackage Nightly](http://stackage.org/package/log-warper/badge/nightly)](http://stackage.org/nightly/package/log-warper)
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
```
gromakovsky: AFAIU, you changed MIT to MPL.

log-warper-core/src/Log/Core/Class.hs
```diff
@@ -0,0 +1,229 @@
...
+withR = withLogAction . cmap
+
+{- | Simple implementation for 'MonadLogger' type class. This data type just
+stores 'LogAction' which performs logging action under monad @m@.
+-}
+newtype LoggerT msg m a = LoggerT
+    { runLoggerT :: ReaderT (LogAction m msg) m a
+    } deriving (Functor, Applicative, Monad, MonadIO)
+
+-- TODO: depend on mtl and implement MonadError and etc?..
```
gromakovsky: s/and etc/etc

log-warper/CHANGES.md
```diff
@@ -0,0 +1,268 @@
+2.0
+===
+
+* Completely rewrite `log-warper`. It's not split into two packages:
```
gromakovsky: s/not/now?

log-warper/src/Log/Actions/Rotation.hs
```diff
@@ -0,0 +1,4 @@
+module Log.Actions.Rotation
+       ) where
```
gromakovsky: wtf

log-warper/log-warper.cabal
```diff
@@ -0,0 +1,181 @@
...
+extra-source-files:  CHANGES.md
+                   , README.md
+                   , examples/HowTo.md
+                   , examples/HowTo.lhs
+                   , examples/PureLogging.md
+                   , examples/PureLogging.lhs
+stability:           experimental
+build-type:          Simple
+cabal-version:       >=2.0
+description:         This package implements nice and featureful wrapper around hslogger library.
```
gromakovsky: AFAIU, it's not a wrapper around `hslogger` library anymore.

log-warper/src/Log.hs
```diff
@@ -0,0 +1,24 @@
+-- | Reexports all functions from @Log.*@ including @Log.Core@ from
+-- @log-warper-core@ package. See README and tutorials for details.
+
+module Log
```
gromakovsky: I am not sure whether `Log` is a good name (too general and doesn't associate with `log-warper`), but ok.

log-warper/src/Log/Actions/Terminal.hs
```diff
@@ -0,0 +1,21 @@
...
+import Log.Core.Action (LogAction (..), cmap)
+import Log.Event (Event, makeRich)
+import Log.Fmt (richEventF)
+
+-- | Very simple action which just prints 'Text' to @stdout@.
+logPrint :: MonadIO m => LogAction m Text
+logPrint = LogAction putTextLn
+
+-- | Simple action which prints every incoming event into @stdout@.
+logStdout :: MonadIO m => LogAction m (Event exts)
```
gromakovsky: Names seem confusing. Both functions print to `stdout`, the difference between them is how events are formatted (AFAIU), so I think the difference between names should reflect that.

log-warper/src/Log/Fmt.hs
```diff
@@ -0,0 +1,93 @@
...
+        Nothing  -> ""
+        Just tid -> ":" +| show @String tid |+ ""
+
+    timeF :: Text
+    timeF = case _reTime of
+        Nothing   -> ""
+        Just time -> "[" +| centiUtcTimeF time |+ "] "
+
+-- | Formats UTC time in next format: @"%Y-%m-%d %H:%M:%S%Q %Z"@
+-- but @%Q@ part show only in centiseconds (always 2 digits).
```
gromakovsky: Will number of digits become configurable at some point?

gromakovsky:
> There's v1 branch which is log-warper snapshot

Should be protected, I think.

log-warper-core/src/Log/Core/Class.hs
gromakovsky: I am not sure this class should be part of `core`. There are many different approaches to monadic effects and using a type class for a particular action is only one possible option. There are many other options, e. g. the simplest one is explicit passing of `LogAction` to functions which need it. Another option is `HasLogAction` as you said.
Some people don't need this class, but I suppose `core` is intended to contain definitions which (almost) everybody needs in order to use `log-warper`.

README.md
volhovm: By the way, why did you? I've just looked at the difference between these two and I still think MIT suits better.

log-warper/README.md
```diff
@@ -0,0 +1,104 @@
+log-warper
+==========
+
+[![Build Status](https://travis-ci.org/serokell/log-warper.svg)](https://travis-ci.org/serokell/log-warper)
+[![Hackage](https://img.shields.io/hackage/v/log-warper.svg)](https://hackage.haskell.org/package/log-warper)
+[![Stackage LTS](http://stackage.org/package/log-warper/badge/lts)](http://stackage.org/lts/package/log-warper)
+[![Stackage Nightly](http://stackage.org/package/log-warper/badge/nightly)](http://stackage.org/nightly/package/log-warper)
+[![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg)](https://opensource.org/licenses/MIT)
```
volhovm: old MIT here as well.

log-warper/src/Log.hs
volhovm: I also think it should be something different. `Logw` or `WLog` would be better imho.

log-warper/src/Log/Actions/Rotation.hs
volhovm: This PR is still WIP, this module is not included anywhere yet.

volhovm (commented):
@ChShersh 

>I fixed code according to review comments.

I see, nice.

>I propose to think about it [extensible config] in parallel with other changes

Uuugh, well, ok. But it makes sense to me to revise this idea before we actually release the 2.0. I mean, we can merge this particular PR, and implement other things after just to minimize PR complexity. I do not find current solution acceptable.

log-warper/src/Log/Event.hs
chshersh: @avnik That's definitely a thing worth considering. But I 'm not sure about this without benchmarks. It's better to create another issue, investigate this problem, implment benchmarks and rewrite in future it's needed. With structured logging support in future this will be not `Text` but some complex data type.

log-warper/src/Log/Name.hs
chshersh: @avnik This `Vector` is supposed to be created once when performance doesn't matter. But then it will be used for `HashMap` lookups a lot of time for hierarchical logging. I just assumed, that `Vector` is more performant. Though I have another ideas regarding performance improvements. Though, again, requires benchmarks.

README.md
chshersh: @gromakovsky @volhovm Yes, this place need to be changed. RE MIT vs. MPL-2.0. I remember some discussion in #lootbox that MPL-2.0 is better than MIT. Better contact @kirelagin for details.

log-warper/src/Log/Actions/Rotation.hs
chshersh: Yes, it's not implemented. Parts on `log-warper` can be rewritten further step by step.

log-warper/log-warper.cabal
chshersh: @gromakovsky Nice catch. Need to change this.

log-warper/src/Log.hs
chshersh: @volhovm @gromakovsky O don't have strong preferences of this. I just chose `Log` because it's the shortest one. I like `Logw` more than `WLog` though.

log-warper/src/Log/Fmt.hs
chshersh: I don't see any problem with this. Current design doesn't forbid this.

chshersh:
@avnik 

> can we finally drop hslogger derived GPL3 licensed stuff?

When `log-warper` will be completely rewritten in near future there will be no sign of old `hslogger` and GPL3 license as a consequence. `hslogger` code is not used, modules not exported, it's just dead code. But it's convenient to have to rewrite and not to forget some things. There's a lot of useful stuff in `log-warper` at the moment!

chshersh:
@volhovm 

> But it makes sense to me to revise this idea before we actually release the 2.0. I mean, we can merge this particular PR, and implement other things after just to minimize PR complexity. I do not find current solution acceptable.

Merging PR to `master` doesn't mean releasing `2.0` on Hackage. It just opens opportunities for other things to be implemented in parallel.

chshersh:
@gromakovsky 

> Should be protected, I think.

Done.

log-warper/src/Log/Event.hs
volhovm: That's definitely a thing _not_ worth considering for now. Same as the `Vector` question later. Unless we have the real problem with logger performance I suggest to abandon these considerations and concentrate on the API and general design.

avnik: My comment _here_ was not about performance, but about convinience -- so we want log errors from  `Either String Foo` w/o explicit convert.

log-warper/src/Log/Actions/Rotation.hs
gromakovsky: Well, anyway, can you add a `(` or remove this `)` to avoid triggering people?

chshersh:
@volhovm @int-index @gromakovsky @avnik I fixed review comments and made CI pass. I think I can now press `Squash and Merge` button after somebody approves. And then I can create issues for further plans.

volhovm (approved):
Just make sure we all review the code before the release. There is still a lot of places to be refined (unused code, commented pieces etc).

